PR #5, opened 2018-11-26
Similarity API

caiosba:


caiosba:
Also upgrades ES from 6.4.2 to 6.5.1

PR #10, opened 2019-12-11
471 hosam config runtime

huslage:


huslage:
@hfikry92 Please remove the changes to `.gitlab-ci.yml` and squash these commits.

huslage:
Tests are failing because the environment variables aren't set for travis.

huslage:
> Tests are failing because the environment variables aren't set for travis.

I added the variables to the environment in `configurator` and the tests passed. Although the python script failed because the GITHUB_TOKEN variable wasn't JSON. 

We probably need a check to set the `GITHUB_PARSED_TOKEN=$GITHUB_TOKEN` if the python fails.

PR #11, opened 2020-02-26
7847 vframe alegre

infojunkie:
For @caiosba review plz


app/main/controller/image_similarity_controller.py
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      'success': result
+    }
+
+  @api.response(200, 'image similarity successfully queried.')
+  @api.doc('Make an image similarity query')
+  @api.expect(image_similarity_request, validate=True)
+  def get(self):
+    image = ImageModel.from_url(request.json['url'], {})
+    result = self.search_by_phash(image.phash, request.json['threshold'], request.json['context'], 1, 0)
```
caiosba: Why hard-code the limit as 1? If we query an image that is already indexed, only the same image will be returned... and for other cases, only the best match will be returned... is it what we want?

app/main/controller/image_similarity_controller.py
```diff
@@ -0,0 +1,63 @@
...
+@api.route('/')
+class ImageSimilarityResource(Resource):
+  @api.response(200, 'image signature successfully stored in the similarity database.')
+  @api.doc('Store an image signature in the similarity database')
+  @api.expect(image_similarity_request, validate=True)
+  def post(self):
+    result = True
+    image = ImageModel.from_url(request.json['url'], request.json['context'])
+    db.session.add(image)
+    db.session.commit()
```
caiosba: I noticed that when an identical image is sent, an exception is thrown because the SHA 256 already exists in the database. The most important thing is that we make sure that duplicate hashes are not stored, which is already happening... but would be good to handle this case gracefully.

app/main/controller/image_similarity_controller.py
infojunkie: Thanks, removed limit and offset for now.

app/main/controller/image_similarity_controller.py
infojunkie: I am actually not sure we should NOT store duplicate images... Consider the case of the same uploaded image to two different workspaces. We should either: a) store two records, each with its own context, or b) support multi-team contexts.

caiosba: or (c) have our own "unique key" which could be, on this case, hash + team id

infojunkie: We already have a serial id, so can we relax the SHA256 uniqueness for now?

caiosba: I think so... on Check API side, for example, we allow two identical images under the same team (we just de-duplicate on Smooch Bot).

PR #12, opened 2020-03-15
4959 healthcheck

infojunkie:


caiosba:
Looks good to me! Let's just think about a way to be sure that new configs are not missed.

PR #18, opened 2020-04-07
Ticket #8176: Rollback session in case of exception

caiosba:
But still log the exception... I expect that this way the subsequent requests won't fail.

### DIFF
--- a/app/main/controller/image_similarity_controller.py
+++ b/app/main/controller/image_similarity_controller.py
@@ -21,8 +21,12 @@ class ImageSimilarityResource(Resource):
   def post(self):
     result = True
     image = ImageModel.from_url(request.json['url'], request.json['context'])
-    db.session.add(image)
-    db.session.commit()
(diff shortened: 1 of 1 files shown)


app/main/controller/image_similarity_controller.py
```diff
@@ -21,8 +21,12 @@ class ImageSimilarityResource(Resource):
   def post(self):
     result = True
     image = ImageModel.from_url(request.json['url'], request.json['context'])
-    db.session.add(image)
-    db.session.commit()
+    try:
```
infojunkie, line 24: are you sure this still logs the exception? doesn't `try` absorb it?

caiosba: that's why there is a `raise` in the `except` block

infojunkie: ohhhh I didn't notice it

PR #19, opened 2020-04-08
Initial proposal for re-working DocSim into potentially shared redis-accessible model

DGaffney:


infojunkie:
@caiosba do you have comments on the code or approach here?

caiosba:
> @caiosba do you have comments on the code or approach here?

:lgtm: - I had some comments before regarding parameters but I'm happy to see that they are all addressed now.

Question: does this approach also scales to host big models like for example the Google News 300 (https://github.com/mmihaltz/word2vec-GoogleNews-vectors)?

infojunkie:
Thanks Caio - I'll work on integrating the bigger model to see how it behaves. 

Logistically, I would like to reach a point where these models are pre-downloaded on our S3 (and locally, in MinIO) and read from over there, instead of the ad-hoc way in which the models are currently accessed. This will hopefully allow us to have an easier path to add new models.

PR #21, opened 2020-05-21
add optional parameter for using elasticsearch fuzzy matching

DGaffney:


app/main/controller/similarity_controller.py
```diff
@@ -69,12 +69,13 @@ def get(self):
                     }
                 },
             ]
-
             # FIXME: `analyzer` and `minimum_should_match` don't play well together.
             if 'language' in request.json:
                 conditions[0]['match']['content']['analyzer'] = language_to_analyzer(request.json['language'])
                 del conditions[0]['match']['content']['minimum_should_match']
-
+            if 'fuzzy' in request.json:
```
infojunkie, line 76: Not sure if we need to make this optional, or if the goal is to find the "best" ES search setup for our use cases. If this is just an experiment to see if fuzzy matching performs better than whatever we had previously, then let's reach a conclusion and incorporate fuzzy matching in the search formulation unconditionnally, or drop it altogether. Otherwise, we run the risk of re-exposing all the ES search options in this API, which would increase complexity and the bugs.

PR #23, opened 2020-06-10
Plugin refactor more vsms

infojunkie:


infojunkie:
is this branch still current?

DGaffney:
@infojunkie not any more I think - we at one time wanted this functionality but after vetting FuzzyWuzzy's performance it's just not worth the cost. We may add SBERT and LASER models back in but that would be in a separate PR.

PR #26, opened 2020-08-04
Updates to add Multi SBERT

DGaffney:


app/main/lib/math_helpers.py
```diff
@@ -0,0 +1,34 @@
+import math
+import numpy as np
+import scipy
+def similarity_for_model_name(language_model_name, vecA, vecB):
```
infojunkie, line 4: Now that the `SharedModel` has a `similarity` overridable method, why do we need this?

Also, is `DocSim` still present in the code?

DGaffney: I'll review in the morning - I think you're right on both parts here

PR #28, opened 2020-09-03
Replace URL so s3 internal URL

DGaffney:


DGaffney:
Merging this code, as the block is now entirely in ElasticSearch's capacity for using our similarity plugin and not directly in this update on the repo.

PR #36, opened 2020-10-19
Modified Redis Health Check

Better-Boy:
Line - https://github.com/meedan/alegre/blob/develop/app/main/controller/healthcheck_controller.py#L44

`# FIXME Redis database is not being tested by info()`

Why the change from `info()` to `ping()` is necessary?

1. The purpose of healthcheck is to check if the redis cluster/server is up and running. You don't need any other info than `True/Exception`. 
2. By default, redis has 16 databases, which can be addressed by their indexes starting from 0 and ending at 15. If any value other than `None`(defaults to 0) and 0 to 15 are given, `ping()` will raise a `ResponseError: DB Index is out of range` which is handled in the `except` block
3. The return value of `ping()` is `True` if the connection is fine with checking database details. No need to assign `True` explicitly

### DIFF
--- a/app/main/controller/healthcheck_controller.py
+++ b/app/main/controller/healthcheck_controller.py
@@ -41,8 +41,7 @@ class HealthcheckResource(Resource):
     # Redis
     try:
       r = redis.Redis(host=app.config['REDIS_HOST'], port=app.config['REDIS_PORT'], db=app.config['REDIS_DATABASE'])
-      # FIXME Redis database is not being tested by info()
-      r.info()
+      r.ping()
       result['REDIS'] = True
     except Exception as e:
       result['REDIS'] = str(e)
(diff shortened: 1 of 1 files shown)


Better-Boy:
@huslage Can any of you guys review this PR?

caiosba:
@huslage will this change work for the health checker?

Better-Boy:
@caiosba @huslage Any Update on this PR?

caiosba:
@huslage @dmou any of you know if that change is valid for the AWS healthchecker?

sonoransun:
Looking at the change, using r.ping() instead of r.info() seems perfectly appropriate. There is additional information in info(), however, it doesn't look like any of that is necessary - only a server liveness check.

Regarding AWS, the Redis ElasticCache service for both QA and Live, `qa-redis` and `live-redis` respectively, are monitored via AWS, so there is no concern regarding ping() vs. info() in the context of AWS service health.

Better-Boy:
@sonoransun Thank you for reviewing the PR.  

@caiosba Now that the PR is reviewed, can it be approved and merged?

PR #48, opened 2021-01-23
CHECK-176 rewrite as per discussion with Caio

DGaffney:


app/main/controller/similarity_controller.py
```diff
@@ -36,53 +36,26 @@ def get_body_for_request(self):
...
+            result = es.update(
+                id=request.json["doc_id"],
+                body={"doc": body},
+                index=app.config['ELASTICSEARCH_SIMILARITY']
+            )
+        else:
+            result = es.index(
+                body=body,
+                index=app.config['ELASTICSEARCH_SIMILARITY']
+            )
```
caiosba, line 58: If I'm not mistaken, the logic of this block is:

```
IF doc_id is passed
  update document with that doc_id
ELSE
  create a new document with random id
```

While what we want is:

```
IF doc_id is passed
  IF document with doc_id exists
    update document with that doc_id
  ELSE
    create document with that doc_id
ELSE
  create document with random id
```

Right?

DGaffney: yep! Good catch. Bad me, programming after my usual hours

app/main/controller/image_similarity_controller.py
```diff
@@ -11,30 +11,37 @@
...
 
   @api.response(200, 'image signature successfully stored in the similarity database.')
   @api.doc('Store an image signature in the similarity database')
   @api.expect(image_similarity_request, validate=True)
   def post(self):
-    image = ImageModel.from_url(request.json['url'], request.json['context'])
+    if request.json.get("doc_id"):
+        delete_record(request.json)
+    image = ImageModel.from_url(request.json['url'], request.json.get('doc_id'), request.json['context'])
     self.save(image)
```
caiosba, line 45: Let's be sure that the `DELETE` and `INSERT` run inside the same transaction.

app/main/controller/image_similarity_controller.py
```diff
@@ -11,30 +11,37 @@
...
+  'doc_id': fields.String(required=False, description='image ID to constrain uniqueness'),
   'threshold': fields.Float(required=False, default=0.9, description='minimum score to consider, between 0.0 and 1.0 (defaults to 0.9)'),
   'context': JsonObject(required=False, default=[], description='context')
 })
 
 def _after_log(retry_state):
   app.logger.debug("Retrying image similarity...")
 
+def delete_record(params):
+  deleted = db.session.query(ImageModel).filter(ImageModel.doc_id==params['doc_id']).delete()
```
caiosba, line 24: Let's please be sure that `params['doc_id']` is not `None`.

caiosba: Otherwise we can delete a bunch of images :)

DGaffney: Fair! Updated.

PR #56, opened 2021-02-19
Fix alegre-sbert deployments in QA

sonoransun:
This fix updates the GitLab-CI deployment process to also deploy an updated alegre-sbert task definition. Resolves incident https://meedan.app.opsgenie.com/incident/detail/b3e83537-3c46-4d0f-97ae-28f0da1879ad.


caiosba (approved):
Looks good! Can we also do it for live already, so we're ready when the time comes? Or you want to be sure it first works on QA, @sonoransun ?

PR #73, opened 2021-04-27
Ticket CHECK-372: Skip google image classification if credentials are…

danielafeitosa:
… missing


DGaffney (commented):
Looks mostly fine to me - only thing I'd advise is adding a `raise` error in `image_classification.py` in lieu of the print statement on line 15.

PR #99, opened 2021-06-16
Fix/688 prevent exceptions

amoedoamorim:


DGaffney (approved):
Looks great, thanks for the additional test!

PR #101, opened 2021-06-23
Deploy video model

sonoransun:
This change deploys the new video similarity model for Alegre into the QA and Live environments.


DGaffney (changes requested):
Looks like you used the name `video_model` for the machine - we'll need to pass MODEL_NAME of `video_model` instead of `video` here. Checking QA as it stands right now, `video_model` works, `video` does not. Easier to roll ahead with `video_model` as it's also more specific.

PR #120, opened 2021-07-08
Test/671 contract test

danielevalverde:


DGaffney (approved):
Looks good to me! Just read through and I can't see a location where a variable that was previously set will now be undefined. Pattern looks uncontroversial and should work fine!

PR #192, opened 2022-01-18
Runtime resilience

sonoransun:
This change run the model and worker entry points repeatedly once started, thus avoiding excessive "Essential Task in Container Exited" alarms from CloudWatch.


DGaffney (approved):
Looks great! Simple solution to remove at least one component of the issue here.

PR #220, opened 2022-03-23
Deploy new mdebertav3filipino model

sonoransun:
This change adds support for new `mdebertav3filipino` model in GitLab deployments.


sonoransun:
Will definitely remember to squash this one :P

PR #246, opened 2022-07-22
Change Alegre port to 3100 to avoid conflict on Mac Monterey

hartsick:
Port 5000, which Alegre currently runs on, is now used by AirPlay on
Macs running Monterey. As a result, there is an error that port is in
use when our application tries to use that port in development.

To fix this, I modified the external port to 3100, which it seems
to have been at some point in the past (reflected by Check Readme). For
internal consistency, I went ahead and updated the internal port
to 5000, as well, even though it wasn't really necessary.

Fixes CHECK-2147
Related to https://github.com/meedan/check/pull/69


huslage (approved):
LGTM

PR #248, opened 2022-08-01
Use community wheel of Tensorflow that works with M1

hartsick:
The TensorFlow binary downloaded from a normal TensorFlow 2.3.1 pip install (from requirements)
was crashing when we used the linux/x86_64 emulated arch with M1 macs (which is needed
because TensorFlow does not yet have an arm-supported version).

To solve this, we are using a community wheel of Tensorflow 2.3.1 compiled as we need it.

More on this here: https://github.com/tensorflow/tensorflow/issues/52845

Paired with @ahmednasserswe! We confirmed working on M1 Macs, would appreciate if others could confirm this approach works locally for non-M1 Mac machines (we're hardcoding the TensorFlow version).

CHECK-2147


hartsick:
Devin confirmed he can build and spin up the container on his non-M1 Mac! Merging

PR #251, opened 2022-08-07
Fixes creating text graphs

caiosba:
When I was trying to generate text clusters locally, it didn’t fail, but no clusters were returned. It worked well for images. Looks like some changes to text similarity were not reflects in the graph writer. Looks like "model" should now be "models" and "text" should be "content". I'm not sure, so I'll ask Devin to review it.

Fixes CHECK-2212.


DGaffney:
Looked through the code and this is correct -we had a regression on graph writing. Looks good to me, thank you!

PR #269, opened 2022-11-11
Meedan 2690 remove vectors from response

DGaffney:


app/main/lib/text_similarity.py
```diff
@@ -119,6 +119,13 @@ def insert_model_into_response(hits, model_key):
             hit["_source"]["model"] = model_key
     return hits
 
+def strip_vectors(results):
+    for result in results:
+        vector_keys = [key for key in result["_source"].keys() if key[:7] == "vector_"]
+        for key in vector_keys:
+            result.pop(key, None)
```
computermacgyver, line 126: I changed `result.pop(key,None)` to `del result["_source"][key]` and this worked for me in an offline thing. The changes are make the the `results` object that is passed into the function (i.e., the `return` isn't strictly necessary as the changes affect the original parameter passed in by reference) @DGaffney

DGaffney: Ah there's definitely one bug that it needs to be `result["_source"].pop` (or `del`, either one)- the return that I'm doing is just a convention I prefer to avoid relying on side-effecting results (even though, yes, it's side-effecting)

PR #285, opened 2023-03-09
Cv2 1002 replace p hash with pdq in alegre

DGaffney:


DGaffney:
@ahmednasserswe looking mostly good here but we need a fix for `ModuleNotFoundError: No module named 'pdqhashing'` which is where the tests are failing right now - maybe add to requirements.txt or set up some build to make sure its available at runtime?

app/main/lib/image_hash.py
```diff
@@ -20,14 +26,41 @@ def compute_ahash(im):
...
   """
   return imagehash.phash(ensure_pil(im))
 
+def compute_pdq(im):
+  """Compute perceptual hash using ImageHash library
+    :param im: Numpy.ndarray
+    :returns: Imagehash.ImageHash
+  """
+  with io.BytesIO() as output:
+    ensure_pil(im).save(output, format="GIF")
```
computermacgyver, line 42: Let's either guess the format based on the file extension in the URL , or easier, let's save as TIF format into the buffer, which will not have any compression artefacts/degradation

app/main/lib/image_hash.py
```diff
@@ -20,14 +26,41 @@ def compute_ahash(im):
...
+    pdq_hasher = PDQHasher()
+    hash_and_qual = pdq_hasher.fromBufferedImage(output)
+    print("hash_and_qual")
```
computermacgyver, line 45: Let's comment all of these `print` statements. Anything really important for logging should get logged with the `Logger`

app/main/lib/image_similarity.py
```diff
@@ -145,3 +146,46 @@ def search_by_phash(phash, threshold, context, limit=None):
...
+    matches = db.session.execute(text(cmd), dict(**{
+      'pdq': pdq,
+      'threshold': threshold,
+      'limit': limit,
+    }, **context_hash)).fetchall()
+    keys = ('id', 'sha256', 'pdq', 'url', 'context', 'score')
+    rows = []
+    for values in matches:
+      row = dict(zip(keys, values))
+      row["model"] = "image"
```
computermacgyver, line 186: What do you think about setting `model` equal to `image/pdq` to explicitly differentiate it from the previous phash results @DGaffney @ahmednasserswe ?

migrations/versions/e495509fad52_add_pdq_field.py
```diff
@@ -0,0 +1,40 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('images', sa.Column('pdq', postgresql.BIT(length=256), nullable=True))
+    op.alter_column('images', 'phash',
+               existing_type=sa.BIGINT(),
+               nullable=True)
+    op.create_index(op.f('ix_images_pdq'), 'images', ['pdq'], unique=False)
+    op.alter_column('images', 'phash', nullable=True)
+
```
computermacgyver, line 27: It looks like there are two `alter_column` operations for the `phash` column. Do we need them both or can one of them be removed?

computermacgyver:
For the pdqhash/ing module, I see two options:
1. We can add [this package](https://github.com/faustomorales/pdqhash-python) to `requirements.txt` and use it, or
2. We can add a `setup.py` file to `threatexchange/pdq/python` to install the files already there. 

The symbolic links are not building because the filesystem is case sensitive on Linux and they are not translating correctly. In general, I think either option above will be more stable than symbolic links in any case. Option (1) is using C bindings and thus should be a bit faster than option 2, which is pure python, but it  may be a bit more fiddly to use the C bindings. I'm trying out option 1 now.

computermacgyver:
Package builds but there is a run-time error about binary incompatibilities. It looks like `pdqhash` builds with a different version of numpy than we are using. 
```
alegre-queue_worker-1   | Traceback (most recent call last):
alegre-queue_worker-1   |   File "manage.py", line 15, in <module>
alegre-queue_worker-1   |     from app import blueprint
alegre-queue_worker-1   |   File "/app/app/__init__.py", line 9, in <module>
alegre-queue_worker-1   |     from .main.controller.similarity_controller import api as similarity_ns
alegre-queue_worker-1   |   File "/app/app/main/controller/similarity_controller.py", line 8, in <module>
alegre-queue_worker-1   |     from app.main.lib import similarity
alegre-queue_worker-1   |   File "/app/app/main/lib/similarity.py", line 5, in <module>
alegre-queue_worker-1   |     from app.main.lib.image_similarity import add_image, delete_image, search_image
alegre-queue_worker-1   |   File "/app/app/main/lib/image_similarity.py", line 3, in <module>
alegre-queue_worker-1   |     from app.main.model.image import ImageModel
alegre-queue_worker-1   |   File "/app/app/main/model/__init__.py", line 3, in <module>
alegre-queue_worker-1   |     from .image import Image
alegre-queue_worker-1   |   File "/app/app/main/model/image.py", line 9, in <module>
alegre-queue_worker-1   |     from app.main.lib.image_hash import compute_phash_int, sha256_stream, compute_pdq
alegre-queue_worker-1   |   File "/app/app/main/lib/image_hash.py", line 14, in <module>
alegre-queue_worker-1   |     import pdqhash
alegre-queue_worker-1   |   File "/usr/local/lib/python3.7/site-packages/pdqhash/__init__.py", line 1, in <module>
alegre-queue_worker-1   |     from .bindings import compute, compute_dihedral, compute_float
alegre-queue_worker-1   |   File "pdqhash/bindings.pyx", line 1, in init pdqhash.bindings
alegre-queue_worker-1   | ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

app/main/lib/image_hash.py
computermacgyver: I've commented these now.

computermacgyver:
I've added a `setup.py` to the pure Python version and use that in the Docker file now.

app/main/lib/image_hash.py
computermacgyver: I've updated this to TIF but it still needs to be tested

app/main/lib/image_similarity.py
DGaffney: That's a cheap thing to do that may disentangle a ton of issues to come in the future. I think we should do it.

migrations/versions/e495509fad52_add_pdq_field.py
DGaffney: you could probably drop the last one that just makes it nullable.

computermacgyver:
I've pushed some commits to the branch that
* have us always compute phash and pdq and store both to the database on POST
* introduce a IMAGE_MODEL configuration variable that decides which model to use for GET (defaults to phash)

The goal is that when we do a first merge and deploy there is no change in the observable behavior, but that we start computing both pdq and phash values for all new images. While in that state, we can also run the migration @ahmednasserswe is working on to create pdq hashes for existing/old images. Once all that is done, we can "flip the switch" and start using PDQ for querying (GET)

app/main/lib/image_similarity.py
```diff
@@ -32,6 +32,8 @@ def save(image):
     existing = db.session.query(ImageModel).filter(ImageModel.url==image.url).one()
     existing.context.append(image.context)
     flag_modified(existing, 'context')
+    existing.pdq.append(image.pdq)
+    flag_modified(existing, 'pdq')
```
computermacgyver, line 36: Let's remove these lines and create a different function specifically for updating the pdq hash

app/main/lib/image_similarity.py
```diff
@@ -47,11 +49,30 @@ def save(image):
     db.session.rollback()
     raise e
 
+
+@tenacity.retry(wait=tenacity.wait_fixed(0.5), stop=tenacity.stop_after_delay(5), after=_after_log)
+def alter_pdq(image):
+  try:
+    # First locate existing image and append new context
+    existing = db.session.query(ImageModel).filter(ImageModel.url==image.url).one()
```
computermacgyver, line 57: We should be looking up based on the id, not URL. Remove the code around image.context (we don't want to alter that).

app/main/lib/image_similarity.py
```diff
@@ -47,11 +49,30 @@ def save(image):
...
+    existing.pdq = image.pdq
+    flag_modified(existing, 'pdq')
+    db.session.commit()
+  except NoResultFound as e:
+    # Otherwise, add new image, but with context as an array
+    if image.context:
+      image.context = [image.context]
+    db.session.add(image)
```
computermacgyver, line 65: We should log the case that we fail to find the image, but we should not be (re-)adding as this will potentially create duplicates.

app/main/lib/image_similarity.py
```diff
@@ -60,6 +81,23 @@ def add_image(save_params):
     db.session.rollback()
     raise e
 
+def add_image_pdq(save_params):
+  try:
+    if save_params.get("doc_id"):
+      print("ha")
+      delete_image(save_params)
```
computermacgyver, line 88: don't delete existing images. We just want to update the pdq column in place

app/main/lib/image_similarity.py
```diff
@@ -60,6 +81,23 @@ def add_image(save_params):
...
+    image = ImageModel.from_url(save_params['url'], save_params.get('doc_id'), save_params['context'], save_params.get("created_at"))
```
computermacgyver, line 89: don't create a new ImageModel object, populate from the DB and only calculate and update pdq

app/main/lib/reindex_images.py
```diff
@@ -0,0 +1,46 @@
...
+
+from sqlalchemy import text
+from sqlalchemy.orm.attributes import flag_modified
+from sqlalchemy.orm.exc import NoResultFound
+
+app.app_context()
+def _after_log(retry_state):
+  app.logger.debug("Retrying image similarity...")
+
+def get_all_images():
```
computermacgyver, line 18: Only get images `WHERE pdq IS NULL` . Add a `LIMIT` parameter or we'll exhaust RAM

app/main/lib/reindex_images.py
```diff
@@ -0,0 +1,46 @@
...
+  try:
+    cmd = """
+      SELECT id, sha256, url, context,created_at,pdq  FROM images
+    """
+    matches = db.session.execute(text(cmd)).fetchall()
+    keys = ('id', 'sha256', 'url', 'context','created_at','pdq')
+    rows = []
+    for values in matches:
+      row = dict(zip(keys, values))
+      if values['pdq'] is not None:
```
computermacgyver, line 28: This logic is backwards. We want rows where pdq IS NULL (this is the opposite). As noted above, however, the filter should be in the SQL query, which means we won't need this line.

PR #287, opened 2023-04-03
Support OpenSearch query search on vector fields

sonoransun:
As discussed in DEVOPS-253 there is a slight change in syntax required for script queries against vector fields in OpenSearch. Where as ElasticSearch queries could reference vector fields implicitly in a doc, for OpenSearch `doc[]` must be specified for vector fields.

This change also includes the index definitions for OpenSearch `alegre_similarity` index with and without knn-approximate indexing on vector fields for future use.


huslage:
@sonoransun @DGaffney What's failing here?

PR #289, opened 2023-04-12
Remove deprecated vector_512 fields from Alegre similarity index.

sonoransun:
Per discussion on DEVOPS-253, deprecate the vector_512 fields which are no longer used.


sonoransun:
@DGaffney your fix worked :) Thank you sir!

PR #292, opened 2023-04-24
Using an archive repository since Debian 9/stretch moved to archive.debian.org

caiosba:
As per message here: https://lists.debian.org/debian-devel-announce/2023/03/msg00006.html

```
Hi,

the stretch, stretch-debug and stretch-proposed-updates suites have now
also been imported on archive.debian.org. People still interested in
these should update their sources.list.

I plan to remove the suites from the main archive in about a month
(2023-04-23 or later).

The stretch-backports, stretch-backports-sloppy and related debug suites
will likely move soon as well and might be removed from the main archive
around the same time.

Ansgar
```

The fix here is to point the repositories to archive.debian.org.

Fixes CV2-3046.

### DIFF
--- a/postgres/Dockerfile
+++ b/postgres/Dockerfile
@@ -4,9 +4,10 @@ MAINTAINER Alexey Kovrizhkin <lekovr+dopos@gmail.com>
 
 ENV DOCKERFILE_VERSION 190701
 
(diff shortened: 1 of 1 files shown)


DGaffney:
Thank you so much @caiosba for your generous contribution!

PR #297, opened 2023-05-02
Cv2 3002 support openai ada embeddings in alegre

ahmednasserswe:


app/main/lib/text_similarity.py
```diff
@@ -71,7 +76,7 @@ def get_elasticsearch_base_conditions(search_params, clause_count, threshold):
             'match': {
               'content': {
                   'query': truncate_query(search_params['content'], clause_count),
-                  'minimum_should_match': str(int(round(threshold * 100))) + '%'
+                  'minimum_should_match': str(int(round(float(threshold) * 100))) + '%'
```
computermacgyver, line 79: Note to other code reviewers that this was due to a separate issue where `round(threshold)` failed when `threshold` was of type string.

PR #298, opened 2023-05-02
Adding prefix `openai-` to OpenAI's models' model-key, adding `vector…

ahmednasserswe:
…_openai-text-embedding-ada-002` index to opensearch, and fixing `requirements.txt` to include prerequisites of `openai[embeddings]`


DGaffney (commented):
`prefix_openai = "openai-"` seems like a potentially shaky way to roll out this setup... but I also don't think that a top-level distinct `SharedModel` class would be a better alternative. I think the problem I have is that with all other `SharedModel` classes, we treat them as entirely independent (i.e. the name completely shards off to a separate class of code) - here, we're really introducing this prefix idea that only applies to one type of vector model. It feels like this could come bite us later. Are we really using lots of different OpenAI models? Or just one or two? Should we add a bit more structure to this conditional?

Curious to hear @computermacgyver's thoughts.

app/main/config.py
```diff
@@ -42,7 +42,7 @@ class Config:
     VIDEO_MODEL_L1_SCORE = 0.7
   AUDIO_MODEL = os.getenv('AUDIO_MODEL', 'audio-model')
   IMAGE_MODEL = os.getenv('IMAGE_MODEL', default='phash')
-
+  OPENAI_API_KEY = os.getenv('OPENAI_API_KEY', default=None)
```
skyemeedan, line 45: I think it would be helpful to add a comment with the SSM key for the secret so that it is clear what needs to be added to the environment if it is missing

app/main/lib/text_similarity.py
```diff
@@ -201,3 +208,8 @@ def search_text_by_model(search_params):
     return {
         'result': response
     }
+
+def retrieve_openai_embeddings(text, model_key):
+    openai.api_key = app.config['OPENAI_API_KEY']
```
skyemeedan, line 213: i'm more familiar with `app.config.OPENAI_API_KEY` style syntax for referencing public class variables. (because then the linter will tell me if I get the key wrong! :-)

skyemeedan: If this is making an external API call at this point, maybe worth logging here so we have a timestamp to refer to in debugging (maybe openai utils allready do that)

requirements.txt
```diff
@@ -151,4 +154,3 @@ Werkzeug==0.16.0
 wrapt==1.11.1
 zipp==3.5.0
 pyacoustid==1.2.2
```
skyemeedan, line 156: I think the sentry imports are still required (Devin just added them) and it looks like missing sentry imports are failing the CI?

skyemeedan (changes requested):
really cool to see the OpenAI embeddings running!   I agree that ideally we should come up with a standardized way of validating model keys mapping various keys to appropriate models, but that could happen in the future.

skyemeedan:
I have not pulled the branch to try running it locally, but happy to do that if helpful

app/main/lib/text_similarity.py
```diff
@@ -202,3 +211,8 @@ def search_text_by_model(search_params):
     return {
         'result': response
     }
+def retrieve_openai_embeddings(text, model_key):
+    openai.api_key = app.config['OPENAI_API_KEY']
+    app.logger.info(f"Calling OpenAI API")
+    model_key_without_openai_prefix = model_key[len("openai-"):]
```
computermacgyver, line 217: Use the constant that you'll have defined at the top of the file.

app/main/config.py
ahmednasserswe: Scott have elaborated it to me. And we added 
```# For local development, you can populate this key from SSM.
# The key is /local/alegre/OPENAI_API_KEY```
to the env_file.example

File: app/main/lib/text_similarity.py [hunk 5]
```diff
@@ -202,3 +210,8 @@ def search_text_by_model(search_params):
     return {
         'result': response
     }
+def retrieve_openai_embeddings(text, model_key):
+    openai.api_key = app.config['OPENAI_API_KEY']
+    app.logger.info(f"Calling OpenAI API")
+    model_key_without_openai_prefix = model_key[len(PREFIX_OPENAI):]
+    return openai.embeddings_utils.get_embedding(text, engine=model_key_without_openai_prefix)
```
computermacgyver, line 217: @ahmednasserswe We should handle the case where this line fails. E.g., if the API is down, API key is invalid, etc. 
It should fail gracefully, log the error, and other models should continue.

skyemeedan (approved):
Approving so we can test in QA.  I'm still not quite clear what happens if we can't get openAI embeddings.  Is the idea that it is better to continue because we will at least write the ES indexes?  Clearly we don't want to bring down the service, but my worry would be this might be failing silently for a while, we might not notice the error logs, and a lot of data could get inserted without proper index.

app/main/lib/openai.py
```diff
@@ -0,0 +1,28 @@
...
+PREFIX_OPENAI = "openai-"
+EMBEDDING_CACHE_DEFAULT = 60*60*24*7 #7 days (value is in seconds)
+def retrieve_openai_embeddings(text, model_key):
+
+    r = redis.Redis(host=app.config['REDIS_HOST'], port=app.config['REDIS_PORT'], db=app.config['REDIS_DATABASE'])
+    text=text.strip()
+    textmd5=hashlib.md5(text.encode('utf-8')).hexdigest()
+    key = f"cache_{model_key}_{textmd5}"
+    val_from_cache=r.get(key)
+    if val_from_cache!=None:
```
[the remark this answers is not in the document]
skyemeedan: I think fixing these style things can be helpful, because there won't be so many climate errors (or other syntax warnings) that can mask other stuff

app/main/lib/openai.py
```diff
@@ -0,0 +1,28 @@
...
+        return pickle.loads(val_from_cache)
+    openai.api_key = app.config['OPENAI_API_KEY']
+    app.logger.info(f"Calling OpenAI API")
+    model_key_without_openai_prefix = model_key[len(PREFIX_OPENAI):]
+    try:
+        embeddings = openai.embeddings_utils.get_embedding(text, engine=model_key_without_openai_prefix)
+        r.set(key,pickle.dumps(embeddings))
+        r.expire(key,EMBEDDING_CACHE_DEFAULT)
+    except Exception as e:
```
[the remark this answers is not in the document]
computermacgyver: Intended.

computermacgyver (approved):

app/main/lib/openai.py
```diff
@@ -0,0 +1,28 @@
+from flask import current_app as app
```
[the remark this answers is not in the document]
ahmednasserswe: doesnt follow the common code style

PR #299, opened 2023-05-02
CV2-3081 add correct env var for sentry to hook into

DGaffney:


skyemeedan (approved):
I feel like I've seen 'local' 'dev' and 'localdev' all used to refer to the system setup on our laptops, which one is standard?

DGaffney:
> I feel like I've seen 'local' 'dev' and 'localdev' all used to refer to the system setup on our laptops, which one is standard?

A meaningful question you ask with no known answer, like the number of stars in the universe

PR #305, opened 2023-05-09
CV2-3111: Mean-tokens was removed from hugging face

computermacgyver:
Now hosting a clone in our huggingface account.

This is all that I think is needed to move from the sentence_transformers repo to our copy of the repo on HuggingFace. Unfortunately I'm not yet able to test because huggingface is also have some general down time. Details on the ticket


DGaffney:
Looks totally fine by me!

PR #306, opened 2023-05-12
CV2-3126: order-image-results-desc

computermacgyver:
This has been in the code since the beginning and unlikely to be a big issue since we reorder results in Check API already, but it could have an effect with a low threshold and fixing the order adhere’s with our larger idea to move order and scoring to Alegre


skyemeedan (approved):
Suggestion: add a function comment about how this works to make error less likely
"Use bit_count_image compiled postgres function to find results that score highly against the requested phash, returning only values above threshold. Result rows are sorted in descending order so that the first item has the best match"

DGaffney:
Ditto on adding a comment as per @skyemeedan's review - otherwise lets do this!

computermacgyver:
> Suggestion: add a function comment about how this works to make error less likely "Use bit_count_image compiled postgres function to find results that score highly against the requested phash, returning only values above threshold. Result rows are sorted in descending order so that the first item has the best match"

Good idea. I've added a comment about the function.

PR #316, opened 2023-06-16
Increase ES default limit; rewrite get_body_for_text_document

computermacgyver:
Rewrite has get_body_for_text_document keep all params unless they are specifically changed.

`get_body_for_text_document` was written to remove parameters unless they appeared specifically on its allow list. This has led to numerous issues in the past when we introduce new parameters and forget to update this function. It was the root issue behind 
* CV2-2805 : [Alegre] `min_es_search` has no effect
* CV2-3164: per_model_threshold for vector models not using the value set in similarity settings 

In this current ticket (CV2-3253) the issue was the default applied in the function, but I'm taking the opportunity to switch the underlying behavior to avoid potential future issues.

### DIFF
--- a/app/main/controller/bulk_similarity_controller.py
+++ b/app/main/controller/bulk_similarity_controller.py
@@ -39,7 +39,7 @@ class BulkSimilarityResource(Resource):
             bodies.append(
                 json_parse_timestamp(
                     get_document_body(
-                        similarity.get_body_for_text_document(document, mode = 'store')
+                        similarity.get_body_for_text_document(document, mode='store')
                     )
                 )
             )
--- a/app/main/controller/bulk_similarity_controller.py
+++ b/app/main/controller/bulk_similarity_controller.py
@@ -39,7 +39,7 @@ class BulkSimilarityResource(Resource):
             bodies.append(
                 json_parse_timestamp(
                     get_document_body(
-                        similarity.get_body_for_text_document(document)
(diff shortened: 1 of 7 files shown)


computermacgyver:
Tested locally and similarity appears to be working fine. This PR changes the default number of results from 20 to 1000 and moves from a `allow list` to a `deny/re-format list` approach for the `get_body_for_text_document`

computermacgyver:
Integration tests running at https://github.com/meedan/check-web/pull/1474

skyemeedan (approved):
Looks like bulk_similarity and bulk_update_similarity are using this, so hopefully they will stay consistent.  Can you update the PR description to mention the problem or issue that this PR is fixing?  (assuming we were dropping some params we needed?)

computermacgyver:
Updated. More details in CV2-3253.

Thinking further on this, I want to separate the code paths for storing a document vs. querying for one. When we store a document, we want to original/current behavior of an allow-list. We don't want random things stored to OpenSearch; we should know all the fields. 

In contrast, when querying a document, we need to broadly allow parameters through to the query function for it to use (things like `per_model_threshold`, `min_es_score`, etc.). These are parameters used for querying but not ones that should be stored in OpenSearch. 

It seems the real root issue here is that we were using `get_body_for_text_document` for storing and querying OpenSearch, and the needs of these two code paths are different.

PR #314, opened 2023-06-15
Use bullseye rather than bookworm due to ffmpeg dependency issue

computermacgyver:

PR #318, opened 2023-06-20
CV2-3263 update to bullseye

DGaffney:
https://meedan.atlassian.net/browse/CV2-3263
> We currently have an issue where alegre (and presto) fail due to an issue with chromaprint’s installation process. This is happening because our ffmpeg version that gets installed on build has updated from version 4 to version 5. In version 5, some of the functions that are used to read files within chromaprint are from ffmpeg’s codebase, and no longer exist as they did in version 4. The fix is to slightly modify a function in chromaprint to update for the latest API. This patch should fix chromaprint installs on Alegre.

> We’ve updated the dockerfile to specify the image as FROM python:3.7-bullseye and that seems to work fine. Of course, I think we use production/Dockerfile in Gitlab, so that’s failing now - building another PR now to solve for that ( ) and then we should be good on Gitlab as well as Travis builds.

### DIFF
--- a/production/Dockerfile
+++ b/production/Dockerfile
@@ -1,6 +1,6 @@
 # alegre
 
-FROM python:3.7
+FROM python:3.7-bullseye
 MAINTAINER sysops@meedan.com
 
 #


skyemeedan:
I've lost the thread of the issue and there isn't any context on this PR.  Seems like we are changing to a different docker container with different pre-installed libraries?

skyemeedan:
Seems like this is the same as https://github.com/meedan/alegre/pull/314?

DGaffney:
> I've lost the thread of the issue and there isn't any context on this PR. Seems like we are changing to a different docker container with different pre-installed libraries?

We've had a discussion about it in slack (https://meedan.slack.com/archives/CGSCKA196/p1687219867730889) but the gist is that we use a different Dockerfile in the production deploy. *Why* we do that is not known to me, but we do. Right now, when I am attempting to deploy, the gitlab builds fail because they build off this Dockerfile, which doesn't have the update from #314, which was the original issue. Hope that clarifies a bit!

skyemeedan (approved):

PR #319, opened 2023-06-20
CV2-3257 add drop-in-replacement for opensearch to replace elasticsearch

DGaffney:


DGaffney:
This should work but needs testing on qa environment next to push it any further - @computermacgyver & @skyemeedan let me know if there's any other objections here!

computermacgyver:
I'm happy on this one if @skyemeedan is happy.

DGaffney:
We can't really test if its working locally afaict, @computermacgyver. Locally, we still use Elasticsearch 7.9.2, so we can't replicate what we'll need to see to be sure. What I can tell you, however, is that it plays nicely when we use the `opensearchpy` library in place of the `elasticsearch` library - if it didn't play nicely, we would have seen tons of tests blowing up from `app/test/test_similarity.py`, which mutates a test DB during the test flow.

skyemeedan (approved):

computermacgyver:
Thanks, @DGaffney . I forgot that detail. Thanks for explaining and merging.

PR #254, opened 2022-09-01
Remove platform dependencies to support deployment to ARM servers

huslage:
We need to support deployment to ARM-based servers and Macs. This is the work to de-platform the code.

PR #320, opened 2023-06-23
[CV2-3224] upgrading alegre versions to try to get builds working

skyemeedan:
bumped python to 3.9, removed chromaprint, some other key stuff commnted


skyemeedan:
Note: there is related work on https://github.com/meedan/alegre/pull/254

PR #323, opened 2023-07-03
CV2-3320 - Do NOT wrap items in `doc`

computermacgyver:
Alegre /bulk/similarity endpoint makes docs unretrievable because it puts them within a `doc` element rather than at the root of the Elasticsearch/Opensearch item

Verified this bug exists on QA with Opensearch and not just in dev with Elasticsearch. 

See https://meedan.atlassian.net/browse/CV2-3320


skyemeedan (approved):
I think we also need to add a unit test that checks the structure of the objects returned by API calls,  and maybe even validation on every API response that we are conforming to an expected schema?

DGaffney (approved):
Thank you for finding this Scott!

PR #324, opened 2023-07-03
CV2-3324 update context setting for sentry errors

DGaffney:


skyemeedan:
@DGaffney can you add some context so I know what to look for in review?

Are we using both pybrake and sentry?

DGaffney:
We are still using pybrake and sentry, yes - we have not initiated a ticket to gut errbit out yet - @caiosba this could become the ticket for that, if you wish.

The context for this ticket is fairly narrow, and represented here: https://meedan.atlassian.net/browse/CV2-3324 - the issue appears to be because, when logging an error in sentry with context being added, there's a `dictionary changed size during Iteration` error (https://meedan.sentry.io/issues/4291563646/?project=4504691019677696&query=is%3Aunresolved&referrer=issue-stream&stream_index=0). That happens when you do something like this in python, which isn't allowed:

```
dd = {"one": 1, "two": 2}
for k in dd:
  dd[k+"blah"] = dd[k]+1
```

The solution is to use a non-deprecated context adder, which is what I am doing now.

caiosba:
@DGaffney , yes, I thought we have done that... I'll file a ticket.

skyemeedan (approved):

PR #325, opened 2023-07-05
CV2-3330 gut errbit

DGaffney:


skyemeedan:
Is there any way we can write a test that this is working? (may be hard if it is sending to a remote log system).  What happens when things run locally, does it still send error messages to Sentry?

DGaffney:
> Is there any way we can write a test that this is working? (may be hard if it is sending to a remote log system). What happens when things run locally, does it still send error messages to Sentry?

If DSN is set as none, Sentry receives no message - so, if your environment defines `sentry_sdk_dsn` as a string, it will attempt to contact Sentry. If its an invalid value, I would assume it'd throw an invalid value error. If its none, it will silently pass by: https://docs.sentry.io/product/sentry-basics/dsn-explainer/.

We can put a test together for this.

computermacgyver (approved):
This looks good to me offhand, and I assume you've tried and succeeded in running it all locally, but if not please do so before merging :smiley:

computermacgyver:
Just saw the comments from @skyemeedan after hitting approve; I agree it would be great to have a test if we can

skyemeedan:
> We can put a test together for this.

Maybe there is already a testing pattern in some of the other code using sentry?  Or we could decide that devops has tested sentry well enough, and just confirm that error messages from this code are appearing in sentry before we close the ticket?

PR #328, opened 2023-07-07
CV2-3374 audio flip threshold

DGaffney:


skyemeedan:
@DGaffney can you add context and ticket?

DGaffney:
> @DGaffney can you add context and ticket?

Sure - the ticket is here - https://meedan.atlassian.net/browse/CV2-3374

The text of the ticket reads as follows:

No idea what the history is for why we’re here again, but we are - audio looks to be flipped the wrong way round (e.g. threshold of 100 means match nothing). We’ll need a reversal in alegre, which is super straightforward. The example you can use to confirm this is by adding the two audio files attached here to any workspace - these are nearly identical files - when you set the threshold to anything other than 0, they will not match.

Note that there are files attached on the JIRA ticket that are not attached here because of constraints from github. Going forward, would it make things easier for you if, for every ticket I work on, I paste in the link and the text of the link for the JIRA ticket? Perhaps that can be automated if so?

skyemeedan:
> would it make things easier for you if, for every ticket I work on, I paste in the link and the text of the link for the JIRA ticket? 
Pasting link: yes please! 😄 
Text: I don't think the full next is usually needed usually (and if the link is there,  reviewers can click through), but at one liner (perhaps pasted from ticket) about why, and some description about how the change fixes the problem, and how it was tested I think is really helpful. 

> Perhaps that can be automated if so?
I think important thing is the description of how the changes fixes the problem, which usually isn't in the original ticket (which may be just a description of the problem). 

> threshold of 100 means match nothing
Are these thresholds in integers in 0-100 range (code looks like expecting floats) 

This seems like it would have been *dramatically* not working, like we would have been hearing complaints from partners?    How will we test that it is working when we roll it out?

DGaffney:
> > would it make things easier for you if, for every ticket I work on, I paste in the link and the text of the link for the JIRA ticket?
> 
> Pasting link: yes please! 😄 Text: I don't think the full next is usually needed usually (and if the link is there, reviewers can click through), but at one liner (perhaps pasted from ticket) about why, and some description about how the change fixes the problem, and how it was tested I think is really helpful.
> 
> > Perhaps that can be automated if so?
> 
> I think important thing is the description of how the changes fixes the problem, which usually isn't in the original ticket (which may be just a description of the problem).
> 
> > threshold of 100 means match nothing
> 
> Are these thresholds in integers in 0-100 range (code looks like expecting floats)
> 
> This seems like it would have been _dramatically_ not working, like we would have been hearing complaints from partners? How will we test that it is working when we roll it out?

No partners are using audio right now, so it snot surprising we didn't observe it.

You're correct - I'm being fast and loose with my language here. A threshold of 1.0 is what I imply by saying 100. 

We will test that it is working with the reproduction steps I outlined in the description I pasted here, and in the original JIRA ticket, along with the files I attached that should match.

DGaffney:
Ok - I think the fix I'm sending in makes sense. Here's the annotated round trip:

Check-API: "Please return audio for ProjectMedia 123 where threshold > 0.75"
Alegre (AudioModel.search function): "Set internal representation of threshold to 1-0.75 (0.25) - if none given, set to 1-0.0 (1.0)"
Alegre (AudioModel.search_by_hash_value function): "Select matches where score <= :threshold, order score ASC" (remember that chromaprint perfect result is 0.0, perfect mismatch is 1.0)
Alegre (AudioModel.search_by_hash_value function): "Reset score that is returned to external providers as 1-score" (to reset as 1 is perfect match, 0 is perfect mismatch)

skyemeedan (approved):

PR #330, opened 2023-07-11
Cv2 3380 log interesting errors

DGaffney:


app/main/controller/audio_transcription_controller.py
```diff
@@ -86,9 +86,9 @@ def get_transcription():
             result = None
             response = self.aws_get_transcription(jobName)
             if response['TranscriptionJob']:
-                job_name = response['TranscriptionJob']['TranscriptionJobName']
-                job_status = response['TranscriptionJob']['TranscriptionJobStatus']
-                language_code = response['TranscriptionJob']['LanguageCode']
+                job_name = response['TranscriptionJob'].get('TranscriptionJobName')
```
skyemeedan, line 89:
```suggestion
                # these properties may be missing if the the recording lacked speech
                # or there were other transcription errors so we let them default to None
                job_name = response['TranscriptionJob'].get('TranscriptionJobName')
```

app/main/controller/audio_transcription_controller.py
```diff
@@ -99,5 +99,8 @@ def get_transcription():
                     transcriptionResponse = requests.get(transcriptionUri)
                     transcriptionResponseDict = json.loads(transcriptionResponse.text)
                     result['transcription'] = transcriptionResponseDict['results']['transcripts'][0]['transcript']
+                if job_status == "FAILED":
```
skyemeedan, line 102: are these the only two job status? (if there was another one we might skip and fail to log?)

skyemeedan (approved):

app/main/controller/audio_transcription_controller.py
DGaffney: fair - no idea. I'll add an else.

PR #336, opened 2023-07-19
CV2-3467 upgrade alegre to psql 13

DGaffney:
## Description
We're bumping up to Postgres 13 - so we need to upgrade the local image to accommodate. There's nothing in the [postgres 13 changelog](https://www.postgresql.org/docs/13/release-13.html) that raises any immediate concerns, but we need to run all these tests to make sure that we're not breaking something.

Reference: CV2-3467

## How has this been tested?
I've run this locally - a few tests broke but I think they were largely because of minor local docker configuration differences. I'd like to see them pass here, and get the code to a point where it passes here, then we'll run an integration test to confirm we're good to go.

### DIFF
--- a/postgres/Dockerfile
+++ b/postgres/Dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:12-buster
+FROM postgres:13-buster
 
 MAINTAINER Alexey Kovrizhkin <lekovr+dopos@gmail.com>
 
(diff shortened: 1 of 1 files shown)


DGaffney:
Talked it through with @caiosba and no full integration test is needed since if it works for alegre, it will work for the rest once the rest are tested. Let's get this accepted but wait for a merge, as we learned our lesson about merging a psql upgrade on alegre before....

computermacgyver (approved):
Thanks. I won't second guess @caiosba's recommendation on integration tests. This looks fine for Alegre but given we've learned that Alegre's file is used for all of Check locally I agree it makes sense to hold on merging until other services have been tested. :pray:

skyemeedan (approved):

vasconsaurus:
confirmed with @DGaffney and @caiosba, I'm merging this one 😁

PR #340, opened 2023-08-31
CV2-3435 add presto endpoint, initial stab at removing adding items from alegre audio

DGaffney:
## Description
The idea here is to add a new callback-reception endpoint. In place of where we currently add audio items, instead, send a post request with the proper message format to Presto. Presto then fires a callback at the new callback-reception endpoint, which then in turn calls the old audio storage code.

Reference: 3435

## How has this been tested?
Updated and gutted tests that directly stored audio objects via POST requests to /audio/similarity. In place of that, we just check that we can hit the stub for the new Presto module, which in turn is just a formatted post request out to Presto. Also, adding a test that checks that we can POST to /presto/receive/add_item/audio to store an audio given a realistic Presto message format. We'll continue that sort of style up test updating as we roll out fixes across presto.


app/main/controller/presto_controller.py
```diff
@@ -0,0 +1,20 @@
...
+    'body': JsonObject(required=False, description='Original body of the message sent - will contain ID, callback_url, and original url/text of request'),
+    'response': JsonObject(required=False, description='The response data sent back from the model'),
+})
+@api.route('/receive/<string:action>/<string:model_type>')
+class PrestoResource(Resource):
+    @api.response(200, 'Successfully responded to presto callback.')
+    @api.doc('Receive a presto callback for a given `model_type`')
+    def post(self, action, model_type):
+        if action == "add":
+            return similarity.add_item(request.args or request.json, model_type)
```
computermacgyver, line 18: I think this may be replying to Presto and not the original caller we want to reply to (e.g., check API). I suspect we need to have some sort of internal redis queue as we do with the shared model servers to receive the callback and enqueue the result back to the original process that is handling the request from Check API unless we change Check API to a callback architecture as well

skyemeedan: discussion today was on changing "everything" to callbacks: https://docs.google.com/document/d/1kIZSy2rQ7npCo_u5NPfyoFoEgLdB6vyJ--mnDxn9SDc/edit#heading=h.hndo1trhkuyf

app/main/controller/presto_controller.py
```diff
@@ -0,0 +1,23 @@
...
+    'body': JsonObject(required=False, description='Original body of the message sent - will contain ID, callback_url, and original url/text of request'),
+    'response': JsonObject(required=False, description='The response data sent back from the model'),
+})
+@api.route('/receive/<string:action>/<string:model_type>')
+class PrestoResource(Resource):
+    @api.response(200, 'Successfully responded to presto callback.')
+    @api.doc('Receive a presto callback for a given `model_type`')
+    def post(self, action, model_type):
+        data = request.args or request.json
+        if action == "add_item":
```
skyemeedan, line 18: lets add delete also, if we if throw 'not implement' for now

app/main/controller/presto_controller.py
```diff
@@ -0,0 +1,23 @@
...
+from app.main.lib.fields import JsonObject
+from app.main.lib import similarity
+
+api = Namespace('presto', description='presto callback url')
+presto = api.model('presto', {
+    'body': JsonObject(required=False, description='Original body of the message sent - will contain ID, callback_url, and original url/text of request'),
+    'response': JsonObject(required=False, description='The response data sent back from the model'),
+})
+@api.route('/receive/<string:action>/<string:model_type>')
+class PrestoResource(Resource):
```
skyemeedan, line 13: maybe this could be 'callback_controller', since other services (like timpani) would use this also?

app/main/lib/presto.py
```diff
@@ -0,0 +1,22 @@
+import uuid
+from flask import current_app as app
+import requests
+
+class Presto:
```
skyemeedan, line 5: ... or is the idea that each service using alegre would have its own set of callback endpoints?

app/main/lib/shared_models/audio_model.py
```diff
@@ -212,14 +202,15 @@ def get_context_for_search(self, task):
...
 
     def search(self, task):
-        audio, temporary = self.get_audio(task)
-        context = self.get_context_for_search(task)
+        body = task.get("body", {})
+        audio, temporary = self.get_audio(body)
+        context = self.get_context_for_search(body)
         if audio:
-            threshold = task.get('threshold', 0.0)
+            threshold = task.get("raw", {}).get('threshold', 0.0)
```
skyemeedan, line 209: looks like we are making commitments that "raw" has  `threshold` and `limit` params, I think this should be enforced somewhere

app/main/lib/similarity.py
```diff
@@ -93,6 +104,10 @@ def add_item(item, similarity_type):
   app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for add was {response}")
   return response
 
+def callback_add_item(item, similarity_type):
+  if similarity_type == "audio":
+      return audio_model().add(item)
```
skyemeedan, line 109: probably nothing is listening to this response, should we log it?

app/main/lib/similarity.py
```diff
@@ -119,3 +135,7 @@ def get_similar_items(item, similarity_type):
   app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for search was {response}")
   return response
 
+def callback_get_similar_items(item, similarity_type):
```
skyemeedan, line 138: is item here the url for the audio item,  or the fingerprint vector we are searching with?

app/main/controller/presto_controller.py
DGaffney: I don't know if this is the right direction to take this code. If I were to implement this right now, I'd want to define a `similarity.callback_delete_item` method, which just goes right back to the delete command that would (and should) otherwise use the existing functionality (which basically is just adding another set of unnecessary code right now). If I don't do that, and raise a not implemented error, I'd just end up with functionally identical code:

```
    def post(self, action, model_type):
        if action == "add_item":
            return similarity.callback_add_item(data, model_type)
        elif action == "search_item":
            return similarity.callback_get_similar_items(data, model_type)
        elif action == "delete_item":
            abort(404, description=f"Action type of {action} was not found. Currently available action types are add_item, search_item.")
        else:
            abort(404, description=f"Action type of {action} was not found. Currently available action types are add_item, search_item.")
```
VS:

```
    def post(self, action, model_type):
        if action == "add_item":
            return similarity.callback_add_item(data, model_type)
        elif action == "search_item":
            return similarity.callback_get_similar_items(data, model_type)
        else:
            abort(404, description=f"Action type of {action} was not found. Currently available action types are add_item, search_item.")
```

I think this may be some overengineering at this point of development. If it's ok with you, I would really prefer we hold on designing more in this direction for now?

app/main/lib/presto.py
DGaffney: I think for now, lets leave it as presto, and when we need to genericize it because we have two callbackable services, we genericize it then as a subclass setup.

app/main/lib/similarity.py
DGaffney: At this point in the code, we have the vector from presto in hand now

app/main/lib/similarity.py
DGaffney: Fair enough!

app/main/lib/shared_models/audio_model.py
DGaffney: I don't know how much we want to enforce this stuff - if we *do* want to enforce it, it's way way upstream of here, at similarity.py/model_response_package function - that's where this initially gets set.

app/main/controller/presto_controller.py
DGaffney: See my other comment w/r/t service-based callbacks

caiosba:
@DGaffney , I'm going to review later today, but one question: does it keep backwards compatibility with the current Check API? Just to see if we can really merge after review is done or we need to wait for the changes on Check API side too.

app/main/lib/similarity.py
skyemeedan: Ok, lets make that clear.  maybe arg is `item_vector`?  validate assumptions about structure before passing to audio model? (i.e. if item really is a numeric vector?)

app/__init__.py
```diff
@@ -44,3 +45,4 @@
 api.add_namespace(image_ocr_ns, path='/image/ocr')
 api.add_namespace(article_ns, path='/article')
 api.add_namespace(graph_ns, path='/graph/cluster')
+api.add_namespace(presto_ns, path='/presto')
```
caiosba, line 48: This is the callback endpoint that Presto calls?

DGaffney: Yep!

DGaffney:
> @DGaffney , I'm going to review later today, but one question: does it keep backwards compatibility with the current Check API? Just to see if we can really merge after review is done or we need to wait for the changes on Check API side too.

Yes this is backwards compatible for everything *other* than audio. This should not affect any flows for any other similarity measures (except the modality we are directly changing right now of course). We'll test a bunch in QA to confirm!

app/main/lib/similarity.py
DGaffney: I'm tossing in an assertion routine - please take a look at it in the latest push to this PR and let me know if this satisfies what you're looking for.

app/main/controller/presto_controller.py
```diff
@@ -0,0 +1,23 @@
...
+    @api.response(200, 'Successfully responded to presto callback.')
+    @api.doc('Receive a presto callback for a given `model_type`')
+    def post(self, action, model_type):
+        data = request.args or request.json
+        if action == "add_item":
+            return similarity.callback_add_item(data, model_type)
+        elif action == "search_item":
+            return similarity.callback_get_similar_items(data, model_type)
```
computermacgyver, line 21: This is still replying in the "wrong" place as far as I can see @DGaffney . I believe this function is called by Presto returning the audio hash. If Alegre replies here, isn't it replying to Presto (rather than replying to the original caller, e.g., Check API)?

DGaffney: Marking as outdated now but if you're still seeing an issue here let me know! I think I've resolved the problem you had raised...

Dockerfile
```diff
@@ -1,39 +1,47 @@
...
+COPY ./threatexchange/tmk/cpp /app/threatexchange/tmk/cpp
+COPY ./threatexchange/pdq/cpp /app/threatexchange/pdq/cpp
 RUN make -C /app/threatexchange/tmk/cpp
+
+# Copy chromaprint files and build
+COPY ./chromaprint /app/chromaprint
 RUN cd chromaprint && cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TOOLS=ON .
 RUN cd chromaprint && make
 RUN cd chromaprint && make install
+
```
computermacgyver, line 19: Should we be removing chromaprint here since the fingerprints will now be computed by Presto?

app/main/controller/presto_controller.py
computermacgyver: I had the same question. Will look for the other comment :-)

app/main/controller/presto_controller.py
computermacgyver: agreed that the generic `else` clause can capture this.

app/main/controller/presto_controller.py
```diff
@@ -0,0 +1,24 @@
...
+})
+@api.route('/receive/<string:action>/<string:model_type>')
+class PrestoResource(Resource):
+    @api.response(200, 'Successfully responded to presto callback.')
+    @api.doc('Receive a presto callback for a given `model_type`')
+    def post(self, action, model_type):
+        data = request.args or request.json
+        if action == "add_item":
+            result = similarity.callback_add_item(data, model_type)
+            CheckAPI.return_storage_webhook(app.config['CHECK_API_HOST'], action, model_type, result)
```
computermacgyver, line 21: This makes me feel that this is a `checkapi_callback` endpoint not a `presto` one. In general it feels problematic to hardcode a callback to CheckAPI here. If Timpani wants to store an item we should be using all the same logic and just getting the callback URL from the initial call. Either that callback URL should be passed along the whole time or cached in Reddis by Alegre and looked up here.

app/main/lib/similarity.py
```diff
@@ -80,8 +90,9 @@ def model_response_package(item, command):
 
 def add_item(item, similarity_type):
   app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] Adding item")
+  callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], similarity_type)
   if similarity_type == "audio":
-    response = audio_model().get_shared_model_response(model_response_package(item, "add"))
+    response = Presto.send_request(app.config['PRESTO_HOST'], "audio__Model", callback_url, model_response_package(item, "add"))
```
computermacgyver, line 95: Should this use `PRESTO_MODEL_MAP` and not have `audio__Model` hardcoded?

app/main/lib/similarity.py
```diff
@@ -93,6 +104,12 @@ def add_item(item, similarity_type):
   app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for add was {response}")
   return response
 
+def callback_add_item(item, similarity_type):
+  if similarity_type == "audio":
+      response = audio_model().add(item)
+  app.logger.info(f"[Alegre Similarity] CallbackAddItem: [Item {item}, Similarity type: {similarity_type}] Response looks like {response}")
+  return response
```
computermacgyver, line 111: add an `else` and log an error whenever `similarity_type` is anything other than `audio`

app/main/lib/similarity.py
```diff
@@ -119,3 +136,12 @@ def get_similar_items(item, similarity_type):
   app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for search was {response}")
   return response
 
+def validate_item(item, similarity_type):
```
computermacgyver, line 139: I don't see this function called anywhere in the diff. Where do we use it? Are `assert` statements sufficient or do we need to log?

app/main/model/audio.py
```diff
@@ -6,19 +6,10 @@
 import numpy as np
 from pydub import AudioSegment #requires ffmpeg and ffprobe to be on the PATH
```
computermacgyver, line 7: The `AudioSegment` import appears to be unused. Remove?

computermacgyver: We may also then be able to remove `pydub==0.25.1` from `requirements.txt`

app/main/lib/similarity.py
```diff
@@ -80,8 +90,9 @@ def model_response_package(item, command):
 
 def add_item(item, similarity_type):
```
computermacgyver, line 91: Feels like we could have an optional callback URL here rather than assuming the callback is always CheckAPI. We'd then store that callback in Reddis with a unique id before sending the request to Presto (with that unique id)

computermacgyver (commented):
I've put a few notes throughout. I'm also wondering where we document general changes in API exceptions. AFAIK, the search method for audio used to allow an audio file to be passed by URL as the query. It appears that is no longer possible. Given this is a breaking API change it feels like we should explicitly test for cases where search is called with a URL and log these.

app/main/lib/shared_models/audio_model.py
```diff
@@ -100,7 +89,8 @@ def delete(self, task):
 
     def add(self, task):
         try:
-            audio = Audio.from_url(task.get("url"), task.get("doc_id"), task.get("context", {}))
+            body = task.get("body", {})
+            audio = Audio.from_url(body.get("url"), body.get("raw", {}).get("doc_id"), body.get("context", {}), body.get("response", {}).get("hash_value"))
```
computermacgyver, line 93: What happens when `hash_value` is `None` here and `url` and `doc_id` are not already in the DB? 
* Do we throw an error?
* Does it continue and run a search that returns nothing?
* Does it continue and run a search that returns everything?

app/main/lib/check_api.py
```diff
@@ -0,0 +1,14 @@
...
+class CheckAPI:
+    @staticmethod
+    def return_storage_webhook(check_api_host, action, model_type, data):
+        check_api_token = app.config['CHECK_API_TOKEN']
+        request = requests.post(f"{check_api_host}/api/webhooks/alegre?token={check_api_token}", json={
+            "action": action,
+            "model_type": model_type,
+            "data": data,
+        })
+        return request
```
computermacgyver, line 14: Per my comment above, it feels like this should be a URL that Check API passes to Alegre when calling the API and not something Alegre should come up with on its own. I'm fine with it as a short term measure, however.

app/main/lib/shared_models/audio_model.py
```diff
@@ -212,14 +207,28 @@ def get_context_for_search(self, task):
...
+            threshold = body.get('threshold', 0.0)
+            limit = body.get("limit")
+        audio, temporary = self.get_audio(body)
+        context = self.get_context_for_search(body)
+        if audio.chromaprint_fingerprint is None:
+            callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], "audio")
+            if task.get("doc_id") is None:
+                task["doc_id"] = str(uuid.uuid4())
+            response = json.loads(Presto.send_request(app.config['PRESTO_HOST'], "audio__Model", callback_url, task).text)
+            result = Presto.blocked_response(response, "audio")
```
computermacgyver, line 225: Do we have to block here? I was thinking/hoping we can return a generic response to the caller at this point like: ```{"result": "pending", "doc_id": f"task['doc_id']"}```
When Presto callsback, the callback handler will recall search and `audio.chromaprint_fingerprint` will no longer be `None`

app/main/lib/shared_models/audio_model.py
```diff
@@ -212,14 +207,28 @@ def get_context_for_search(self, task):
...
+            context = self.get_context_for_search(result["body"]["raw"])
         if audio:
-            threshold = task.get('threshold', 0.0)
             matches = self.search_by_hash_value(audio.chromaprint_fingerprint, threshold, context)
-            limit = task.get("limit")
             if temporary:
-                self.delete(task)
+                self.delete(body)
             if limit:
                 return {"result": matches[:limit]}
```
computermacgyver, line 233: If we take my comment above, then this line and the return below as well need to:
1. Call the callback URL to Check API if we have previously returned / called Presto
2. Simply return as we do if we have not.

computermacgyver (commented):
I've added two comments that I think if implemented would remove the need for `Presto.blocked_response`. Overall, very excited about how this is going!

app/main/lib/shared_models/audio_model.py
computermacgyver: I agree with the general point here. We've gotten bitten multiple times by issues of parameters being passed / not passed correctly. That said, I think we should open a separate ticket and address it independently of this work. Therefore I'm going to resolve this conversation for now.

app/main/lib/shared_models/audio_model.py
DGaffney: I believe this would continue and run a search that returns nothing - how would you like to handle @computermacgyver ?

app/main/lib/shared_models/audio_model.py
DGaffney: I think lets just block - this is easier to deal with for now IMO

app/main/lib/check_api.py
DGaffney: The only problem with doing that is that we have to persist and plumb the URL that Check API would pass through *every* request (alegre->presto, presto->alegre), or come up with a persistence mechanism. I think this is reasonable in the short term as, inside the data, there will always be an unambiguous ID that Check API-side can unpack to determine state, *and*, that URL that is being used won't ever change on the Check API side anyways (i.e. that is always going to be the endpoint that receives this sort of thing, no matter what)

caiosba (commented):
Looks good overall! I'm not going to explicitly approve/reject the PR since Scott and Skye are already on this. I just have one question: why the controllers/constants with name Check API? I didn't see anything here that was specific for Check API.

app/main/lib/shared_models/audio_model.py
computermacgyver: I'm fine to block here, but I believe that means the HTTP request between Check API and Alegre will be held open for the duration of this block. We may get an API Gateway Timeout if the roundtrip exceeds one minute. Having the block here means we're not using the Check API callback URL (Check API experiences a blocking call). I'm OK with that, but I felt we wanted to avoid it wherever possible. It shouldn't be too much to avoid the blocking call here, but we can leave that as future work.

app/main/lib/similarity.py
computermacgyver: bump. I think we should still log an error here when `similarity_type` is not `audio`

app/main/lib/check_api.py
computermacgyver: Agreed, but there are two longer term issues. 
(1) We ultimately need Alegre to work with Check API and Timpani. This means the callback URL will be different depending on the service calling Alegre. 
(2) Alegre shouldn't know anything about its callers. To date, Alegre has no knowledge/specifics about what Check API is or even that it exists. It feels like best practice for a microservice to keep Alegre agnostic to any specifics of its caller. This goes to the question that @caiosba asked about the controllers/constants for Check API.

app/main/lib/shared_models/audio_model.py
computermacgyver: I think we can leave as is. I want to test manually when I'm running the code.

app/main/lib/check_api.py
caiosba: @DGaffney and I discussed that yesterday and I think we agreed to rename the constants/controller to something generic like "callback".

PR #342, opened 2023-09-18
Adding more detailed debugging output 

skyemeedan:
adding more detailed error debugging output in the response when alegre is unable to find the requested model on lookup (perhaps because skye forgot to start it in the local dev ..)

Note: my editor is applying an opinionated formatter, which has adjusted whitespace to python standards.  The actual change is on line 64, adding:

```
 assert model_info is not None, f"Unable locate model info for key {model_key}"
```

## Description
Please include a very brief high-level description of the problem or feature and technical details or specific code changes on PR

Reference: TICKET-ID (to provide additional context)

## How has this been tested?
Has it been tested locally? 
I confirmed that the alegre system ran locally and gave the desired error response

Are there automated tests?
none added

This does return additional detail on an error, which can leak info about internal state of the system, however this is an internal service API

### DIFF
--- a/app/main/lib/shared_models/shared_model.py
+++ b/app/main/lib/shared_models/shared_model.py
@@ -12,141 +12,159 @@ import re
 
 from flask import current_app as app
 
-Task = namedtuple('Task', 'task_id task_type task_package')
+Task = namedtuple("Task", "task_id task_type task_package")
+
 
 class SharedModel(object):
-  @staticmethod
-  def import_model_class(model_class):
-    class_name = re.sub(r'(?<!^)(?=[A-Z])', '_', model_class).lower()
-    return getattr(importlib.import_module('app.main.lib.shared_models.%s' % class_name), model_class)
-
-  @staticmethod
-  def start_server(model_class, model_key, options={}):
-    class_ = SharedModel.import_model_class(model_class)
-    instance = class_(model_key, options)
-    instance.load()
-    app.logger.info('[%s] Serving model...', model_key)
-    r = redis.Redis(host=app.config['REDIS_HOST'], port=app.config['REDIS_PORT'], db=app.config['REDIS_DATABASE'])
-    r.set('SharedModel:%s' % model_key, json.dumps({
-      'model_class': model_class,
-      'model_key': model_key,
-      'options': options
-    }))
-    r.sadd('SharedModel', model_key)
-    instance.bulk_run()
-
-  @staticmethod
-  def get_servers():
-    r = redis.Redis(host=app.config['REDIS_HOST'], port=app.config['REDIS_PORT'], db=app.config['REDIS_DATABASE'])
-    return [server.decode('utf-8') for server in r.smembers('SharedModel')]
-
-  @staticmethod
-  def get_client(model_key, options={}):
-    r = redis.Redis(host=app.config['REDIS_HOST'], port=app.config['REDIS_PORT'], db=app.config['REDIS_DATABASE'])
-    model_class = json.loads(r.get('SharedModel:%s' % model_key).decode('utf-8'))['model_class']
-    class_ = SharedModel.import_model_class(model_class)
-    return class_(model_key, options)
-
-  def __init__(self, model_key, options={}):
-    self.options = options
-    self.queue_name = model_key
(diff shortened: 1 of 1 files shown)


app/main/lib/shared_models/shared_model.py
```diff
@@ -12,141 +12,159 @@
...
-  def run(self):
-    while True:
-      task = self.get_task()
-      if task:
-        self.send_response(
-          task,
-          self.respond(task.task_package)
+    @staticmethod
+    def import_model_class(model_class):
+        class_name = re.sub(r"(?<!^)(?=[A-Z])", "_", model_class).lower()
```
DGaffney, line 21: What would be the source of a model class name where it could even potentially fall afoul of this list? Is it free text entry at some point upstream of here?

app/main/lib/shared_models/shared_model.py
```diff
@@ -12,141 +12,159 @@
...
+        r = redis.Redis(
+            host=app.config["REDIS_HOST"],
+            port=app.config["REDIS_PORT"],
+            db=app.config["REDIS_DATABASE"],
+        )
+        return [server.decode("utf-8") for server in r.smembers("SharedModel")]
+
+    @staticmethod
+    def get_client(model_key, options={}):
+        r = redis.Redis(
```
DGaffney, line 58: You're instantiating like this over and over. Maybe worth having a RedisServer module or something?

app/main/lib/shared_models/shared_model.py
```diff
@@ -12,141 +12,159 @@
...
+        model_info = r.get("SharedModel:%s" % model_key)
+        assert model_info is not None, f"Unable locate model info for key {model_key}"
+        model_class = json.loads(model_info.decode("utf-8"))["model_class"]
+        class_ = SharedModel.import_model_class(model_class)
+        return class_(model_key, options)
+
+    def __init__(self, model_key, options={}):
+        self.options = options
+        self.queue_name = model_key
+        self.datastore = redis.Redis(
```
DGaffney, line 72: Again, there's a ton of redis.Redis in this file - feels like this could be DRY'ed up a bit

app/main/lib/shared_models/shared_model.py
```diff
@@ -12,141 +12,159 @@
...
-      tasks = self.get_tasks()
-      responses = ([self.respond(task.task_package) for task in tasks])
-      for i, response in enumerate(responses):
-        self.send_response(
-          tasks[i],
-          response
+    def get_task(self, timeout=0):
+        item = self.datastore.blpop(self.queue_name, timeout)
+        if item:
+            task = Task(**json.loads(item[1].decode("utf-8")))
```
DGaffney, line 81: Worth using pydantic validation on the task data maybe?

DGaffney (changes requested):
I know I'm making some comments about code that pre-existed you jumping in here but may be worth addressing some of them while you're here?

skyemeedan:
This PR is intended a one line change to give more descriptive errors back to the caller that I needed timpani debugging.  I could open a ticket for more general cleanup of shared_model code and move these comments there?

DGaffney:
> This PR is intended a one line change to give more descriptive errors back to the caller that I needed timpani debugging. I could open a ticket for more general cleanup of shared_model code and move these comments there?

Seems reasonable - @computermacgyver what do you think?

computermacgyver (approved):
Agree. The whitespace changes make it difficult to spot the single change, but checking `model_info` is not none before trying to use it (and giving a useful error message when it is) is a good change.

computermacgyver:
The items you've noted @DGaffney are good points --- let's not those in new JIRA tickets. I think we don't need to address them in this PR, however, since they are preexisting and unchanged except for whitespace.

computermacgyver (changes requested):
Let's please make this two separate PRs. We should make the one line change and merge that without any further whitespace changes. I'm concerned otherwise that git blame will not work well and we'll confuse ourselves in the future about why we edited lines xyz. 

Separately, we could have a different PR that fixes whitespace across the whole project?

computermacgyver:
Was planning to open a clean PR but accidentally committed to `develop` rather than the new branch. I'm inclined to leave that unless anyone has issues. Sorry. Need to wake up!
https://github.com/meedan/alegre/commit/bd14839af5cdf5ffd0ec0cede214da61130e027e

PR #346, opened 2023-10-03
Update version pins for Python dependencies to fix deployments.

sonoransun:
## Description

This fix updates version pins for python packages awscli, boto3, botocore, and ecs-deploy to fix deployments.

Reference: https://meedan.atlassian.net/browse/DEVOPS-502

## How has this been tested?

I verified this build and deployment process in QA.

Yes :)


sonoransun:
> This looks great (and thank you). Was it just the CI tests that ran, or did you try running tests locally? (I had trouble with version escalation when trying to bump some versions flagged by dependabot, but I don't think it was these specific dependencies)

I only tried CI and deployments, but to be clear this is version pins for the GitLab deployment stages, rather than Alegre as a whole!

skyemeedan (approved):

PR #349, opened 2023-10-11
Cv2 3827 sync endpoint

DGaffney:
## Description
We are adding a new pair of endpoints for similarity lookup on alegre - one that is explicitly, consistently synchronous, and one that is explicitly, consistently, asynchronous. This is the first of these, the sync endpoint. We're extending existing audio model capacity to afford for a unique flow to explicitly run a blocking search, and calling into the code through that.

Reference: CV2-3827

## How has this been tested?
This has been tested locally - while we are updating tons of tests in the audio test suite, we're aiming to *not* cause any changes to the HTTP interface in the existing audio tests. We're also adding tests for the new sync endpoint to help illustrate the package being sent through to it.

### DIFF
--- a/app/main/controller/presto_controller.py
+++ b/app/main/controller/presto_controller.py
@@ -26,8 +26,9 @@ class PrestoResource(Resource):
         data = request.args or request.json
         app.logger.info(f"PrestoResource {action}")
         if action == "add_item":
-            result = similarity.callback_add_item(data, model_type)
-            Webhook.return_webhook(app.config['CHECK_API_HOST'], action, model_type, result)
+            result = similarity.callback_add_item(dict(**data.get("body"), **data.get("response", {})), model_type)
+            if data.get("body", {}).get("requires_callback"):
+                Webhook.return_webhook(app.config['CHECK_API_HOST'], action, model_type, result)
             r = redis.Redis(
                 host=app.config['REDIS_HOST'],
                 port=app.config['REDIS_PORT'],
--- a/app/main/lib/presto.py
+++ b/app/main/lib/presto.py
@@ -3,6 +3,7 @@ import uuid
 import redis
 from flask import current_app as app
 import requests
+from app.main.lib.serializer import safe_serializer
 
 class Presto:
     @staticmethod
@@ -14,14 +15,18 @@ class Presto:
         return f"{alegre_host}/presto/receive/search_item/{similarity_type}"
 
     @staticmethod
-    def send_request(presto_host, model_key, callback_url, message):
-        return requests.post(f"{presto_host}/process_item/{model_key}", json={
+    def send_request(presto_host, model_key, callback_url, message, requires_callback=True):
+        data = {
             "callback_url": callback_url,
             "id": message.get("doc_id", str(uuid.uuid4())),
             "url": message.get("url"),
             "text": message.get("text"),
-            "raw": message
-        })
+            "raw": message,
+            "requires_callback": requires_callback
+        }
+        headers = {"Content-Type": "application/json"}
+        json_data = json.dumps(data, default=safe_serializer)
+        return requests.post(f"{presto_host}/process_item/{model_key}", data=json_data, headers=headers)
 
     @staticmethod
     def blocked_response(message, model_type):
--- a/app/main/lib/serializer.py
+++ b/app/main/lib/serializer.py
@@ -1,0 +1,16 @@
+import json
+import datetime
+
+def safe_serializer(obj):
+    """
+    Serialize datetime objects into string format.
+    
+    Args:
+    - obj (object): The object to be serialized
+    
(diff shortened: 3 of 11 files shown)


app/main/lib/serializer.py
```diff
@@ -0,0 +1,16 @@
+import json
+import datetime
+
+def safe_serializer(obj):
```
skyemeedan, line 4: `date_serializer`?

app/main/lib/presto.py
```diff
@@ -14,14 +15,18 @@ def get_similar_items_callback_url(alegre_host, similarity_type):
         return f"{alegre_host}/presto/receive/search_item/{similarity_type}"
 
     @staticmethod
-    def send_request(presto_host, model_key, callback_url, message):
-        return requests.post(f"{presto_host}/process_item/{model_key}", json={
+    def send_request(presto_host, model_key, callback_url, message, requires_callback=True):
+        data = {
             "callback_url": callback_url,
```
skyemeedan, line 20: oh, I see the callback url is in the data payload?

app/main/controller/presto_controller.py
```diff
@@ -26,8 +26,9 @@ def post(self, action, model_type):
         data = request.args or request.json
         app.logger.info(f"PrestoResource {action}")
         if action == "add_item":
-            result = similarity.callback_add_item(data, model_type)
-            Webhook.return_webhook(app.config['CHECK_API_HOST'], action, model_type, result)
+            result = similarity.callback_add_item(dict(**data.get("body"), **data.get("response", {})), model_type)
+            if data.get("body", {}).get("requires_callback"):
+                Webhook.return_webhook(app.config['CHECK_API_HOST'], action, model_type, result)
```
skyemeedan, line 31: instead of `requires_callback`, maybe this could just be `callback`  and the caller could set the callback url?  i.e. check would set it as CHECK_API_HOST and timpani would set it to match https://github.com/meedan/timpani/blob/6134e47866fb72a58ecdc0ce36977425abffd728/timpani/conductor/app.py#L111

app/test/test_audio_similarity.py
```diff
@@ -152,17 +152,17 @@ def test_search_by_doc_id(self):
         db.session.commit()
         with patch('app.main.lib.shared_models.audio_model.AudioModel.get_by_doc_id_or_url', ) as mock_get_by_doc_id_or_url:
```
skyemeedan, line 153: isn't the function being mocked here one of the ones that is being tested?

skyemeedan (commented):
still brainstorming endpoint names because, as '_sync' sounds to me like it will synchronize things.  what about `get_similarity` for synchronous and `request_similarity` for async?

app/main/lib/presto.py
DGaffney: Yeah that is what I'm imagining - if we explicitly pass it in, that'll make it much more flexible to other use cases rather than just obliquely implying them with some other structure?

app/test/test_audio_similarity.py
DGaffney: Partially. The goal of this test is not to check whether or not the getting of an individual item works or not, the goal is to make sure that that *rest* of the *search* functionality works, *conditional* on it being a lookup via doc id. If you'd like, I can relax that constraint and rewrite this test.

app/main/controller/presto_controller.py
DGaffney: I'm not sure about that - I think that conceptually disentangling it may end up being useful? I can see cases in my head where we don't want to / wont have the flexibility of deleting the callback URL from the other side of the request, but also want to be able to deny the callback anyways? I think they are two distinct ideas. I am happy to change that if everyone disagrees.

app/main/lib/serializer.py
DGaffney: I went back and forth on that name. I chose `safe` because, where it is used, the json serialize option doesn't care about the fact that this is a specific, narrow serialization, and over time, this would be the singular location to introduce safe serializations by type

app/main/lib/similarity.py
```diff
@@ -139,3 +153,17 @@
...
   return response
+
+def blocking_get_similar_items(item, similarity_type):
+  app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] searching on item")
+  if similarity_type == "audio":
+    response = audio_model().blocking_search(model_response_package(item, "search"))
+  elif similarity_type == "video":
+    response = video_model().blocking_search(model_response_package(item, "search"))
+  elif similarity_type == "image":
+    response = blocking_search_image(item)
```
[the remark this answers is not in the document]
skyemeedan: are these missing functions still a problem?

skyemeedan (approved):

PR #352, opened 2023-10-18
Cv2 3827 update contract

DGaffney:
## Description
Changes to accommodate the sync endpoint accidentally caused a regression on the existing expectations. This fix accommodates both more comprehensively.

Reference: CV2-3827

## How has this been tested?
We're adding new tests to explicitly make sure that this regression can't happen again


skyemeedan:
pairs well with: https://github.com/meedan/check-api/pull/1699

skyemeedan (approved):

PR #354, opened 2023-10-23
CV2-3827 initial push on adding async endpoints WIP

DGaffney:
## Description
Adding an async endpoint which explicitly calls the get_similar_items endpoint in a non-blocking fashion. Mostly just extends existing functionality to a new endpoint. Should go in fine without much editing but we'll see!
Reference: 3827

## How has this been tested?
I've added tests on the async controller as well as the presto callback controller, and have tested this directly from repl. I have not tested this in an integration test with check api yet, as we are not yet using it there. That's likely up next!


computermacgyver:
> I have not tested this in an integration test with check api yet, as we are not yet using it there. That's likely up next!

That's fine. We still want to run integration tests to ensure nothing has changed with the normal endpoints.

app/main/controller/presto_controller.py
```diff
@@ -24,12 +24,16 @@ class PrestoResource(Resource):
     @api.doc('Receive a presto callback for a given `model_type`')
     def post(self, action, model_type):
         data = request.args or request.json
-        print(data)
         app.logger.info(f"PrestoResource {action}")
         if action == "add_item":
+            app.logger.info(f"Data looks like {data}")
```
computermacgyver, line 29: Let's move this before the `if` so that we log the content of `data` in all cases?

app/main/lib/similarity.py
```diff
@@ -164,3 +173,12 @@ def blocking_get_similar_items(item, similarity_type):
...
       raise Exception(f"{similarity_type} modality not implemented for blocking requests!")
 
+def async_get_similar_items(item, similarity_type):
+  app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] searching on item")
+  if similarity_type == "audio":
+    response = audio_model().async_search(model_response_package(item, "search"))
+    app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for search was {response}")
+    return response
+  else:
+      raise Exception(f"{similarity_type} modality not implemented for blocking requests!")
```
computermacgyver, line 183: Minor: Error message should be "not implemented for **callback** requests"

computermacgyver (approved):

PR #355, opened 2023-10-30
Bugfix : move Alegre migrations into their own tasks in ECS

sonoransun:
## Description
This change moves the migration steps in Alegre into their own distinct Makefile target, which in turn is run individually in ECS as a distinct migration task similar to how check-api runs migration.

This will isolate the schema changes / migration steps into their own task with its own logs and prevent issues with migrations from affecting ongoing service performance.

Reference: [DEVOPS-517](https://meedan.atlassian.net/browse/DEVOPS-517)

## How has this been tested?
I ran test migration in QA, however this did not actually make any changes. One thing I would like to discuss is how to test some actual schema and index changes as part of a real migration attempt in QA. I'll discuss with @DGaffney and @computermacgyver , and perhaps @skyemeedan as some of his testing in QA has triggered issues with Alegre.

Yes, this code does not present any additional security concerns.

[DEVOPS-517]: https://meedan.atlassian.net/browse/DEVOPS-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DGaffney:
Looks fine to me - please let me know if you need help getting the tests fixed, otherwise I'm chill with this!

DGaffney (approved):
Approved pending test fix

computermacgyver (changes requested):
We need to update https://github.com/meedan/alegre/blob/bugfix/no-migrations-test/production/bin/start.sh as well.

sonoransun:
> We need to update https://github.com/meedan/alegre/blob/bugfix/no-migrations-test/production/bin/start.sh as well.

Thanks Scott, I should have pulled this request while it undergoes further changes from another branch testing additional logging and initialization changes in the migration steps. As mentioned, we also need the updated OpenSearch definitions here too, and some changes to the parent check repo for bringing up a new alegre service locally (since these migrations are now only called during QA or Live deployments)

PR #356, opened 2023-11-03
CV2-2875 rework response structures

DGaffney:
## Description
In light of more prep for getting ready to build out other modalities, I'm slightly reworking response structures to make them more consistent across modalities - most of the work is on the presto end, but it will require a few tweaks on this side as well.

Reference: 2875

## How has this been tested?
This has been tested by updating fixtures on local unit tests, and running integration tests against this PR to confirm that the updated fixtures actually look the same as the real data coming out of presto. Should be good to go!


skyemeedan (approved):
seems like the change is naming the list 'body' instead of 'response'

PR #357, opened 2023-11-08
Adding nullable column to postgres images table with type Vector to s…

ahmednasserswe:
## Description
Adding nullable column to postgres images table with type Vector to store sscd embedding vectors.
This included changes to both Postgress dockerfile and Alegre dockerfile.

Reference: TICKET-ID  CV2-3917

## How has this been tested?
tested locally 

No potential security threads

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y ffmpeg cmake swig libavcodec-dev libavformat-dev
 RUN ln -s /usr/bin/ffmpeg /usr/local/bin/ffmpeg
 
+#RUN git clone https://github.com/pgvector/pgvector-python.git
+#RUN cd pgvector-python && pip install -r requirements.txt
+
 # Copy necessary threatexchange folders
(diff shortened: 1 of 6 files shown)


skyemeedan:
This looks really promising, probably lots of cool things we can do with direct access to the vector.  But it looks like we need to first get Alegre migrated to python 3.8?  Build is failing with 
```
ERROR: Ignored the following versions that require a different python version: 0.2.0 Requires-Python >=3.8; 0.2.1 Requires-Python >=3.8; 0.2.2 Requires-Python >=3.8; 0.2.3 Requires-Python >=3.8

ERROR: Could not find a version that satisfies the requirement pgvector==0.2.3 (from versions: 0.1.0, 0.1.1, 0.1.2, 0.1.3, 0.1.4, 0.1.5, 0.1.6, 0.1.7, 0.1.8)

ERROR: No matching distribution found for pgvector==0.2.3
```

DGaffney:
Upgrading to 3.8 is a good idea - I'm usually against over-eager version bumps, but this code is now very long in the tooth and 3.8 is very old. 

@skyemeedan I wonder if we should hold on the version bump until we release all the dependencies off alegre? That would very likely make a bump much easier...

PR #358, opened 2023-11-13
CV2-3828 update contract

DGaffney:
## Description
This is a super small update to just change the contract tests to only use body instead of url query params. I've verified that all controllers support both bodies or query params, so we'll sunset query params eventually, but this shouldn't break anything in the interim.

Reference: 3828

## How has this been tested?
This is breaking tests on check-api, so that's how it's being tested right now!

Not particularly relevant here


skyemeedan (approved):
can you make corresponding update to the alegre API docs if we are not using query params anymore?

DGaffney:
> can you make corresponding update to the alegre API docs if we are not using query params anymore?

Happy to!

PR #360, opened 2023-11-28
Keep doc_id when storing in Elastic/Opensearch

computermacgyver:
Current code strips doc_id before storing items in Opensearch/Elastisearch. This is an error and will result in deletes and updates not functioning correctly. Fixing here.

## Description
Please include a very brief high-level description of the problem or feature and technical details or specific code changes on PR

Reference: CV2-4052

## How has this been tested?
Has it been tested locally? Are there automated tests?

Please list any security concerns that may be relevant.

### DIFF
--- a/app/main/lib/similarity.py
+++ b/app/main/lib/similarity.py
@@ -72,7 +72,7 @@ def get_body_for_text_document(params, mode):
       params['content'] = None
 
     if mode == 'store':
-      allow_list = set(['language', 'content', 'created_at', 'models', 'context', 'callback_url'])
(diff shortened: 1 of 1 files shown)


computermacgyver:
I suspect deletes are silently failing at present. I agree we should have a test for this. I guess the question is whether we want to deploy the fix quickly and add the test later or wait until we have both.

skyemeedan:
I think it is worth adding the test, so we are sure this fixes (maybe if it is removed somewhere else also).

PR #363, opened 2023-11-29
start overhauling and simplifying code path for presto fingerprinting by modality

DGaffney:
## Description
This work pushes images to using presto via the sync endpoint, and also starts to streamline and generalize the basic crud operations for media objects (images, videos, audios) since they are nearly identical spare their search functions once presto is integrated fully. 

Reference: 4044

## How has this been tested?
This will break a ton of stuff - will need a bunch of reworked unit tests. This initial push is to just surface all of that.


DGaffney:
@skyemeedan and @computermacgyver can I get one last round of eyes on this now that we're ready to merge?

skyemeedan:
> This will break a ton of stuff - will need a bunch of reworked unit tests. This initial push is to just surface all of that.

Looks like the local tests are passing, is it working with integration tests?

.env_file.test
```diff
@@ -10,6 +10,8 @@ DATABASE_PASS=postgres
 PROVIDER_LANGID=google
 PROVIDER_IMAGE_CLASSIFICATION=google
 ALEGRE_HOST=http://alegre:3100
-PRESTO_HOST=http://presto:8000
+PRESTO_HOST=http://presto-server:8000
```
skyemeedan, line 13: is there a corresponding update to docker-compose file to rename the service?

app/main/lib/image_similarity.py
```diff
@@ -60,47 +61,54 @@ def add_image(save_params):
...
     if context_query:
       cmd = """
-          SELECT id, sha256, phash, url, context FROM images
+          SELECT id, doc_id, phash, url, context FROM images
           WHERE 
         """+context_query
     else:
       cmd = """
-          SELECT id, sha256, phash, url, context FROM images
+          SELECT id, doc_id, phash, url, context FROM images
```
skyemeedan, line 106: has the table structure changed? are we now using doc_id instead of hash to look up images?  Is this compatible with data already in the db?

app/test/test_audio_similarity.py
```diff
@@ -291,34 +291,34 @@ def test_audio_model_confirmed_no_match(self):
...
-
-    def test_handle_http_error(self):
-        with patch('app.main.lib.shared_models.audio_model.AudioModel.save', ) as mock:
-            url = 'file:///app/app/test/data/test_audio_1.mp3'
-            mock.side_effect = urllib.error.HTTPError(url, 420, "HTTP ERROR HAPPENED", None, None)
-            result = self.model.add({"url": url, 'doc_id': "Y2hlY2stcHJvamVjdF9tZWRpYS01NTQ1NzEtdmlkZW8", "context": {"has_custom_id": True}})
-            self.assertIsInstance(result, dict)
-            self.assertEqual(False, result['success'])
-
+    # def test_search_by_context(self):
```
skyemeedan, line 294: are these tests intended to be commented?

app/main/model/image.py
```diff
@@ -30,6 +30,15 @@ class ImageModel(db.Model):
     db.Index('ix_images_context', context, postgresql_using='gin'),
   )
 
+  @classmethod
+  def from_task_data(cls, task):
```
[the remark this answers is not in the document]
skyemeedan: Is this an initializer for a class?  what class?

skyemeedan (approved):
Approving to unblock.  I'm not sure about implications of service name change and switch from sha256 to doc id for lookups. (not very sure how this existing code works)

app/main/lib/image_similarity.py
caiosba: Seconding this... is it a simple change in the column name or is it a different value being stored? Is it going to impact existing data, or new images will still be able to match existing images?

.env_file.test
caiosba: Yes, in the `check` repo.

caiosba:
> > This will break a ton of stuff - will need a bunch of reworked unit tests. This initial push is to just surface all of that.
> 
> Looks like the local tests are passing, is it working with integration tests?

Yes, there was a branch in `check-web` to be sure integration tests were passing.

app/main/lib/image_similarity.py
DGaffney: This is not going to break anything downstream - we just don't need to return the sha because no one uses it, while the doc_id is much more useful

app/test/test_audio_similarity.py
DGaffney: Yep - we no longer serve the audio model directly, so tests that test for that are moot

PR #364, opened 2023-12-04
Bugfix/no migrations test

sonoransun:
## Description
This change migrates Alegre over to OpenSearch with limited backwards compatibility for the legacy model name `elasticsearch` for certain queires.

References:
https://meedan.atlassian.net/browse/DEVOPS-517
https://meedan.atlassian.net/browse/CV2-3255

## How has this been tested?
This has been tested extensively in QA to verify the new migration task process, and the updating opensearch bindings, and the configuration and naming changes.

I am mostly concerned with breaking existing behavior. I tried to cover as much as possible, and will see how automated tests look when triggered for this pull request.


.gitlab-ci.yml
```diff
@@ -104,6 +108,12 @@ deploy_live:
     - pip install ecs-deploy==1.14.0
     - pip install awscli==1.29.59
     - aws ssm get-parameters-by-path --region $AWS_DEFAULT_REGION --path /live/alegre/ --recursive --with-decryption --output text --query "Parameters[].[Name]" | sed -E 's#/live/alegre/##' > env.live.names
+    - for NAME in `cat env.live.names`; do echo -n "-s live-alegre-migration $NAME /live/alegre/$NAME " >> live-alegre-migration.env.args; done
+    - ecs update live-alegre-migration --image live-alegre-migration $QA_ECR_API_BASE_URL:$CI_COMMIT_SHA --exclusive-env -e live-alegre-migration APP alegre -e live-alegre-migration DEPLOY_ENV live -e live-alegre-migration AWS_REGION $AWS_DEFAULT_REGION -e live-alegre-migration ALEGRE_PORT 8000 --exclusive-secrets `cat live-alegre-migration.env.args`
```
skyemeedan, line 112: `$LIVE_ECR_API_BASE_URL` ?

skyemeedan (commented):
Wow, this amazing, huge amount of work!
- do we actually need to change the name of the model? ie 'ELASTICSEARCH_SIMILARITY' could still be used to describe the 'model' (indexing structure) even if it is hosted in OpenSearch?
- If we are changing the name we use to refer to doing a lookup using Open/ElasticSearch full text indexes, maybe we should change it to something like 'FULLTEXT_INDEX' so that it describes the kind of index (vs the hosting technology)

.gitlab-ci.yml
sonoransun: > `$LIVE_ECR_API_BASE_URL` ?

This is one of the few things (live deploy pipeline) that I can't test easily. Thank you!

sonoransun:
> * do we actually need to change the name of the model? ie 'ELASTICSEARCH_SIMILARITY' could still be used to describe the 'model' (indexing structure) even if it is hosted in OpenSearch?

for sanity sake, I wanted to keep things consistent once we migrate. I did wonder about naming this something more generic, like `SIMILARITY_MODEL_PREFIX` due to the way it is used. Hmm...


> * If we are changing the name we use to refer to doing a lookup using Open/ElasticSearch full text indexes, maybe we should change it to something like 'FULLTEXT_INDEX' so that it describes the kind of index (vs the hosting technology)

`FULLTEXT_INDEX` is a better name :) This brings to mind a discussion about how the language specific indices are generated (_fr, _de, _en, etc...) but that's not a refactoring to tackle here...

app/main/controller/similarity_async_controller.py
```diff
@@ -11,18 +11,18 @@
...
     'threshold': fields.Float(required=False, description='minimum score to consider, between 0.0 and 1.0 (defaults to 0.9)'),
     'context': JsonObject(required=True, description='context'),
-    'fuzzy': fields.Boolean(required=False, description='whether or not to use fuzzy search on GET queries (only used when model is set to \'elasticsearch\')'),
+    'fuzzy': fields.Boolean(required=False, description='whether or not to use fuzzy search on GET queries (only used when model is set to \'opensearch\')'),
     'requires_callback': fields.Boolean(required=False, description='whether or not to trigger a callback event to the provided URL'),
 })
 @api.route('/<string:similarity_type>')
 class AsyncSimilarityResource(Resource):
     @api.response(200, 'text similarity successfully queried.')
-    @api.doc('Make a text similarity query. Note that we currently require GET requests with a JSON body rather than embedded params in the URL. You can achieve this via curl -X GET -H "Content-type: application/json" -H "Accept: application/json" -d \'{"text":"Some Text", "threshold": 0.5, "model": "elasticsearch"}\' "http://[ALEGRE_HOST]/text/similarity"')
```
computermacgyver, line 24: This should not change. The model is called `elasticsearch` when interacting with the Alegre API regardless of what we actually have in the backend.

app/main/controller/similarity_controller.py
```diff
@@ -10,12 +10,12 @@
 similarity_request = api.model('similarity_request', {
     'text': fields.String(required=False, description='text to be stored or queried for similarity'),
     'doc_id': fields.String(required=False, description='text ID to constrain uniqueness'),
-    'model': fields.String(required=False, description='similarity model to use: "elasticsearch" (pure Elasticsearch, default) or the key name of an active model'),
-    'models': fields.List(required=False, description='similarity models to use: ["elasticsearch"] (pure Elasticsearch, default) or the key name of an active model', cls_or_instance=fields.String),
+    'model': fields.String(required=False, description='similarity model to use: "opensearch" (pure Elasticsearch, default) or the key name of an active model. Legacy elasticsearch model supported for migration purposes.'),
```
computermacgyver, line 13: Again. This should not change.

app/main/controller/similarity_controller.py
```diff
@@ -10,12 +10,12 @@
...
+    'models': fields.List(required=False, description='similarity models to use: ["opensearch"] (pure Elasticsearch, default) or the key name of an active model. Legacy elasticsearch model supported for migration purposes.', cls_or_instance=fields.String),
     'language': fields.String(required=False, description='language code for the analyzer to use during the similarity query (defaults to standard analyzer)'),
     'threshold': fields.Float(required=False, description='minimum score to consider, between 0.0 and 1.0 (defaults to 0.9)'),
     'context': JsonObject(required=False, description='context'),
-    'fuzzy': fields.Boolean(required=False, description='whether or not to use fuzzy search on GET queries (only used when model is set to \'elasticsearch\')'),
+    'fuzzy': fields.Boolean(required=False, description='whether or not to use fuzzy search on GET queries (only used when model is set to \'opensearch\')'),
```
computermacgyver, line 18: Ditto. Revert to `elasticsearch`

app/main/controller/similarity_controller.py
```diff
@@ -10,12 +10,12 @@
 similarity_request = api.model('similarity_request', {
     'text': fields.String(required=False, description='text to be stored or queried for similarity'),
     'doc_id': fields.String(required=False, description='text ID to constrain uniqueness'),
-    'model': fields.String(required=False, description='similarity model to use: "elasticsearch" (pure Elasticsearch, default) or the key name of an active model'),
-    'models': fields.List(required=False, description='similarity models to use: ["elasticsearch"] (pure Elasticsearch, default) or the key name of an active model', cls_or_instance=fields.String),
+    'model': fields.String(required=False, description='similarity model to use: "opensearch" (pure Elasticsearch, default) or the key name of an active model. Legacy elasticsearch model supported for migration purposes.'),
+    'models': fields.List(required=False, description='similarity models to use: ["opensearch"] (pure Elasticsearch, default) or the key name of an active model. Legacy elasticsearch model supported for migration purposes.', cls_or_instance=fields.String),
```
computermacgyver, line 14: Ditto.

app/main/lib/graph_writer.py
```diff
@@ -58,7 +58,7 @@ def package_item_for_query(item, graph, data_type):
   elif data_type == "text":
     vector_keys = [k for k in item["_source"].keys() if "vector" in k]
     vector_key = ""
-    model = graph.context.get("model") or "elasticsearch"
+    model = graph.context.get("model") or "opensearch"
```
computermacgyver, line 61: Need to revert. This should not change.

app/main/lib/similarity.py
```diff
@@ -45,7 +45,7 @@ def get_body_for_text_document(params, mode):
     if 'models' in params:
         models = models|set(params['models'])
     if not models:
-        models = ['elasticsearch']
+        models = ['opensearch']
```
computermacgyver, line 48: Revert

app/main/lib/text_similarity.py
```diff
@@ -16,7 +16,7 @@ def get_document_body(body):
     context = body.get("context", {})
     if context:
       body["contexts"] = [context]
-    if model_key != 'elasticsearch':
+    if model_key != 'opensearch' and model_key != 'elasticsearch':
```
computermacgyver, line 19: Unnecessary because `model_key` should never be `opensearch` this is an invalid input as far as the Alegre API is concerned..

app/main/lib/text_similarity.py
```diff
@@ -47,10 +47,12 @@ def search_text(search_params):
   return results
 
 def get_model_and_threshold(search_params):
-  model_key = 'elasticsearch'
+  model_key = 'opensearch'
```
computermacgyver, line 50: Revert

app/main/lib/text_similarity.py
```diff
@@ -47,10 +47,12 @@ def search_text(search_params):
...
   threshold = 0.9
   if 'model' in search_params:
       model_key = search_params['model']
+  if model_key == 'elasticsearch':
+      model_key = 'opensearch'
```
computermacgyver, line 55: Need to remove this if block

app/main/lib/text_similarity.py
```diff
@@ -142,12 +144,14 @@ def strip_vectors(results):
...
     out_results = []
+    if model_key == 'elasticsearch':
+        model_key = 'opensearch'
     try:
         min_es_score = float(search_params.get("min_es_score"))
     except (ValueError, TypeError) as e:
         app.logger.info(f"search_params failed on min_es_score for {search_params}, raised error as {e}")
         min_es_score = None
-    if min_es_score is not None and model_key == "elasticsearch":
+    if min_es_score is not None and model_key == "opensearch":
```
computermacgyver, line 154: revert

app/main/lib/text_similarity.py
```diff
@@ -163,17 +167,17 @@ def search_text_by_model(search_params):
...
     clause_count = 0
-    search_indices = [app.config['ELASTICSEARCH_SIMILARITY']]
+    search_indices = [app.config['OPENSEARCH_SIMILARITY']]
     if 'context' in search_params:
         matches, clause_count = generate_matches(search_params['context'])
     if clause_count >= app.config['MAX_CLAUSE_COUNT']:
         return {'error': "Too many clauses specified! Text search will fail if another clause is added. Current clause count: "+str(clause_count)}
-    if model_key.lower() == 'elasticsearch':
-        conditions = get_elasticsearch_base_conditions(search_params, clause_count, threshold)
+    if model_key.lower() == 'opensearch':
```
computermacgyver, line 179: revert

computermacgyver (changes requested):
We need to decide exactly the scope of the changes here. 
* We definitely want to shift to using the Opensearch docker image
* Do we want to rename all the secrets? E.g., change `ELASTICSEARCH_URL` to `OPENSEARCH_URL`? I don't have a strong preference. We just have to make sure not only the `.env_file`s but also the actual secrets in AWS get changed if we do so.
* I do NOT think we should change the Alegre API. The alegre API currently has a model called `elasticsearch` . If we change the name of that model, it will break the Alegre API and all current uses of it across Check, Timpani, etc. We would also need to migrate existing data to rename the model, which is painful. Per my comments in this review, I think the Alegre API should not change. This should be a change internal to Alegre and not something that breaks the API for services external to Alegre.

PR #365, opened 2023-12-06
upgrade redis to 6.2 and specify travis distro

vasconsaurus:
## Description
We had to upgrade Redis in Pender due to the Sidekiq upgrade, and Redis is on 6.2 on our qa and live environment. So we are taking this opportunity to update in all repos.

Regarding the distro, we had issues running the tests on Travis, in Pender changing the disto fixed it. So I'm trying that here as well.

Reference: 3480

## How has this been tested?
Passed alegre's automated tests, and passed integration tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: jammy
 before_install:
 - openssl aes-256-cbc -K $encrypted_e34ab48306dd_key -iv $encrypted_e34ab48306dd_iv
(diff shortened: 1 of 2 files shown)


skyemeedan (approved):
@DGaffney should we merge this and see how it runs in localdev and QA?

DGaffney:
Yep looks good!

PR #367, opened 2023-12-23
Cv2 4150 sscd implement on updated alegre nullable pg vector column in alegre images table

ahmednasserswe:
## Description
Adding SSCD pgvector column to image model

Reference: TICKET-ID Cv2 4150

## How has this been tested?
tested locally

### DIFF
--- a/app/main/model/image.py
+++ b/app/main/model/image.py
@@ -25,7 +25,7 @@ class ImageModel(db.Model):
   doc_id = db.Column(db.String(64, convert_unicode=True), nullable=True, index=True, unique=True)
   phash = db.Column(db.BigInteger, nullable=True, index=True)
   pdq = db.Column(BIT(256), nullable=True, index=True)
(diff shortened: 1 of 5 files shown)


app/main/model/image.py
```diff
@@ -22,14 +25,13 @@ class ImageModel(db.Model):
   doc_id = db.Column(db.String(64, convert_unicode=True), nullable=True, index=True, unique=True)
   phash = db.Column(db.BigInteger, nullable=True, index=True)
   pdq = db.Column(BIT(256), nullable=True, index=True)
-
+  sscd = db.Column(Vector(512), nullable=True, index=True)
```
computermacgyver, line 28: Let's remove `index=True` here @ahmednasserswe . We'll want to stress test on QA and later will likely add an approximate nearest neighbor using HNSW (https://tembo.io/blog/vector-indexes-in-pgvector/)

PR #370, opened 2024-01-10
Cv2 3700 sscd image similarity algorithm

ahmednasserswe:
## Description
Implementing SSCD-copy-detection algorithm. 

Reference: Cv2 3700

## How has this been tested?
Tested locally, to be tested in the PR by automated and integration tests

no security concerns found

### DIFF
--- a/app/main/model/image.py
+++ b/app/main/model/image.py
@@ -25,7 +25,7 @@ class ImageModel(db.Model):
   doc_id = db.Column(db.String(64, convert_unicode=True), nullable=True, index=True, unique=True)
   phash = db.Column(db.BigInteger, nullable=True, index=True)
   pdq = db.Column(BIT(256), nullable=True, index=True)
-  sscd = db.Column(Vector(512), nullable=True, index=True)
+  sscd = db.Column(Vector(512), nullable=True)
   url = db.Column(db.String(255, convert_unicode=True), nullable=False, index=True)
   context = db.Column(JSONB(), default=[], nullable=False)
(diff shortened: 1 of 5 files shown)


computermacgyver (changes requested):
Please pull in the latest from the `develop` branch @ahmednasserswe and re-test

app/main/model/image.py
```diff
@@ -52,11 +54,12 @@ def from_url(url, doc_id, context={}, created_at=None):
     raw = remote_response.read()
     im = Image.open(io.BytesIO(raw)).convert('RGB')
     phash = compute_phash_int(im)
+    sscd = None
```
DGaffney, line 57: I'm not seeing where, if ever, sscd gets set - am I just missing it?

DGaffney (changes requested):
I'm a bit confused as to how SSCD is set - if this is just establishing it as a field then this is fine, otherwise this needs more work.

app/main/model/image.py
computermacgyver: This PR creates the field (nullable) and always set it to null. The next PR should actually set it. My thinking was to break it into two PRs with this first one being "safe" insofar as it just creates the DB column (and always sets it to null)

DGaffney (approved):

computermacgyver (approved):

ahmednasserswe:
Implementing SSCD-copy-detection algorithm.
Adding a nullable column to image of type pgvector to contain the sscd embedding vectors.

Reference: Cv2 3700

PR #373, opened 2024-01-22
CV2-4102 fix a couple more tests that could randomly fail

DGaffney:
## Description
Tests are still a bit flaky - this should sturdy them up!

Reference: CV2-4102

## How has this been tested?
Haven't tested locally but extending pattern already used to address existing flakiness


computermacgyver (approved):
Probably use `assertTrue` rather than `assertEqual(...,True)` but overall changes look great. Thanks!

PR #383, opened 2024-04-01
CV2-4437 stub out article work to prevent illegal instruction crashes

DGaffney:
## Description
Right now, any attempt to import newspaper3k results in an Illegal Instruction exception that hard-kills the python instance. This requires upgrading to newspaper4k, which requires upgrading Python 3.7 to Python 3.8, which requires upgrading tensorflow, which is a pain. We need to eventually pay down that debt - we are now 6 years behind with our code versions in Alegre - this will only become more painful. In lieu of fixing this today, we are opting to simply gut functionality for newspaper-*-based operations on articles, which should unstick us from the current state, which is that no alegre builds are working.

Reference: CV2-4437

## How has this been tested?
Now works locally and starts up - was not starting at all before. Open question if all tests work but I think they should just go through fine.

We need to update our package versions...

### DIFF
--- a/app/main/controller/article_controller.py
+++ b/app/main/controller/article_controller.py
@@ -1,7 +1,7 @@
 import sys
 from flask import abort, request, current_app as app
 from flask_restplus import Resource, Namespace, fields
-from newspaper import Article
+# from newspaper import Article
 from app.main import db
 from app.main.model.article import ArticleModel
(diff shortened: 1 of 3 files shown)


computermacgyver (approved):
Thanks for doing this quickly

PR #384, opened 2024-04-10
CV2-4126 Rebase then continuing video via presto

DGaffney:
## Description
Extends Check-API/Alegre support for Presto to Video modality. Will require S3 downloading capacity! Other than that, should be straightforward.

Reference: 4126

## How has this been tested?
No tests locally yet - will test soon once I start seeing where the code fails

Will need to review S3 access on this...


DGaffney:
@skyemeedan and @computermacgyver this is ready for you to both review!

app/main/lib/media_crud.py
```diff
@@ -126,27 +134,66 @@ def get_context_for_search(task):
...
+        response = get_presto_request_response(modality, callback_url, task)
+        # Warning: this is a blocking hold to wait until we get a response in 
+        # a redis key that we've received something from presto.
+        return obj, temporary, get_context_for_search(task), Presto.blocked_response(response, modality)
+    else:
+        return obj, temporary, get_context_for_search(task), obj.canned_response
 
 def get_async_presto_response(task, model, modality):
-    _, temporary = get_object(task, model)
+    app.logger.error(f"Task at this point is {task}")
```
skyemeedan, line 170: debug?

app/main/lib/redis.py
```diff
@@ -0,0 +1,4 @@
+import redis
```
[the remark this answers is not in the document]
skyemeedan: seems like we have a name conflict with the actual redis package.  maybe the file could be `redis_wrapper.py` or `redis_utils.py`?

app/main/lib/redis.py
```diff
@@ -0,0 +1,4 @@
...
+from flask import current_app as app
```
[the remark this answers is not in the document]
skyemeedan: is there a missing requirement?

app/main/lib/redis.py
```diff
@@ -0,0 +1,4 @@
...
+def get_client():
+    return redis.Redis(host=app.config['REDIS_HOST'], port=app.config['REDIS_PORT'], db=app.config['REDIS_DATABASE'])
```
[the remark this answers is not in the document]
skyemeedan: need to resolve the naming conflict above

app/main/lib/shared_models/audio_model.py
```diff
@@ -56,23 +55,10 @@ def async_search(self, task, modality):
...
-            body["hash_value"] = body.get("result", {}).get("hash_value")
-            body["context"] = body.get("context", body.get("raw", {}).get("context"))
-        else:
-            body = task
-            threshold = body.get('threshold', 0.0)
-            limit = body.get("limit")
+        body, threshold, limit = media_crud.parse_task_search(task)
         audio, temporary = media_crud.get_object(body, Audio)
         if audio.chromaprint_fingerprint is None:
+            app.logger.error(f"Task of {task} and audio of {audio.id} has null chromaprint_fingerprint!")
```
skyemeedan, line 61: seems like this is debug, not error?

app/main/lib/shared_models/video_model.py
```diff
@@ -14,61 +15,96 @@
...
+        s3_client = boto3.client('s3')
+    # Extract the file name from the S3 file path
+    tmk_file = filename.split('/')[-1]
+    # Specify the full path to save the file
+    try:
+        s3_client.download_file(bucket, tmk_file, local_path)
+        app.logger.info(f'Successfully downloaded file {tmk_file} from S3 bucket.')
+    except Exception as e:
+        app.logger.error(f'Failed to download file {tmk_file} from S3 bucket: {e}')
+
```
skyemeedan, line 62: need to re-raise exception here so that the caller will fail correctly?

app/main/lib/shared_models/video_model.py
```diff
@@ -14,61 +15,96 @@
...
 from app.main.lib import media_crud
 from app.main import db
 from app.main.model.video import Video
-FFMPEG_DIR = "/usr/local/bin/ffmpeg"
+from app.main.lib.presto import Presto
+
+MINIO_HOST = "minio:9000"
+def download_file_from_s3(bucket: str, filename: str, local_path: str):
+    """
+    Generic download helper for s3. Could be moved over to helpers folder...
```
skyemeedan, line 35: I agree, this would be better as utility function as not directly a concern of video model

app/main/lib/shared_models/video_model.py
```diff
@@ -14,61 +15,96 @@
...
+        s3_folder = (task.get("result", {}) or {}).get("folder")
+        s3_filepath = (task.get("result", {}) or {}).get("filepath")
+        if hash_value:
+            task["hash_value"] = hash_value
+        added, obj = media_crud.add(task, Video, ["hash_value", "folder", "filepath"])
+        download_file_from_s3(s3_folder, s3_filepath, media_crud.tmk_file_path(obj.folder, obj.filepath))
+        return added
+
+    def blocking_search(self, task, modality):
+        video, temporary, context, presto_result = media_crud.get_blocked_presto_response(task, Video, modality)
```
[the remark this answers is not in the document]
skyemeedan: oh no, we are ignoring context ;-)

app/main/lib/shared_models/video_model.py
```diff
@@ -14,61 +15,96 @@
...
 
     def async_search(self, task, modality):
         return media_crud.get_async_presto_response(task, Video, modality)
 
     def get_tempfile(self):
         return tempfile.NamedTemporaryFile()
 
-    def execute_command(self, command):
-        return os.popen(command).read()
+    def execute_command(self, cmd, params):
```
skyemeedan, line 106: I still don't think we should allow executing arbitrary SQL against the database.  I think is cleaner and safer to have a functions corresponding to the the actions, put the SQL inside like an ORM

app/main/lib/similarity.py
```diff
@@ -176,23 +189,28 @@ def blocking_get_similar_items(item, similarity_type):
...
   if similarity_type == "audio":
-    response = audio_model().async_search(model_response_package(item, "search"), "audio")
+    response, waiting_for_callback = audio_model().async_search(model_response_package(item, "search"), "audio")
     app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for search was {response}")
-    return response
+    return response, waiting_for_callback
   elif similarity_type == "video":
-    response = video_model().async_search(model_response_package(item, "search"), "video")
+    response, waiting_for_callback = video_model().async_search(model_response_package(item, "search"), "video")
+    audio_response, waiting_for_audio_callback = audio_model().async_search(video_model().overload_context_to_denote_content_type(model_response_package(item, "search")), "audio")
```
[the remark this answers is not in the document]
skyemeedan: I think this `audio_model()` search is the the wrong place here in the video block?

app/main/lib/similarity.py
```diff
@@ -176,23 +189,28 @@ def blocking_get_similar_items(item, similarity_type):
...
     app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for search was {response}")
-    return response
+    return response, waiting_for_callback or waiting_for_audio_callback
```
skyemeedan, line 209: response should be video?

app/main/model/audio.py
```diff
@@ -22,11 +24,27 @@ class Audio(db.Model):
     db.Index('ix_audios_context', context, postgresql_using='gin'),
   )
 
+  @property
+  def canned_response(self):
```
skyemeedan, line 28: why is this a "canned_response"? what does it mean? (looks like it is still computing a fingerprint?)

app/main/lib/image_similarity.py
```diff
@@ -63,7 +63,7 @@ def add_image(save_params):
     raise e
 
 def callback_add(task):
-    return media_crud.add(task, ImageModel, ["pdq", "phash"])
+    return media_crud.add(task, ImageModel, ["pdq", "phash"])[0]
```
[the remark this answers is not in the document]
DGaffney: it returns a tuple, and we only want the first item in this context - at least that is what I'm seeing in the code?

app/main/lib/media_crud.py
DGaffney: removing

app/main/lib/shared_models/audio_model.py
DGaffney: another removal - this was a comment during fixes

app/main/lib/shared_models/video_model.py
DGaffney: fair enough!

app/main/lib/shared_models/video_model.py
DGaffney: lets do it!

app/main/lib/shared_models/video_model.py
DGaffney: I understand and am sympathetic to this, but I don't think this is the place to address it. Perhaps lets discuss on a call some time soon to get something on the board? @computermacgyver what do you think?

app/main/lib/similarity.py
DGaffney: This and the audio_model() call are intentional - we need to encode the audio track for all videos to see if we can match them across modes (i.e. this MP3 matches this video's audio track, so they are able to be matched)

app/main/model/audio.py
DGaffney: A canned response is just returning the existing value stored in the DB - if we already have a value for a URL, we are making the assumption that the fingerprint is still valid (which, if we cut the other way, we sign up for re-encoding lots of things over and over needlessly)

app/main/lib/similarity.py
skyemeedan: Oh nice!  Can you add comments to explain?

app/main/model/audio.py
skyemeedan: maybe existing_response?

DGaffney:
@skyemeedan and @computermacgyver one last review on this before we merge if you can?

app/main/lib/similarity.py
```diff
@@ -121,20 +125,31 @@ def callback_add_item(item, similarity_type):
...
+    return {"result": full_responses}
+
 def callback_search_item(item, similarity_type):
   if similarity_type == "audio":
       response = audio_model().search(model_response_package(item.get("raw"), "search"))
       app.logger.info(f"[Alegre Similarity] CallbackSearchItem: [Item {item}, Similarity type: {similarity_type}] Response looks like {response}")
   elif similarity_type == "video":
-      response = video_model().search(model_response_package(item, "search"))
+      video_response = video_model().search(model_response_package(item.get("raw"), "search"))
+      audio_response = audio_model().search(video_model().overload_context_to_denote_content_type(model_response_package(item, "search")))
```
skyemeedan, line 151:
```suggestion
      # video files will (usually?) receive an audio response from transcription of the audio
      audio_response = audio_model().search(video_model().overload_context_to_denote_content_type(model_response_package(item, "search")))
```

skyemeedan (approved):

app/main/lib/similarity.py
DGaffney: This one isn't quite right but I'll add a comment here

PR #390, opened 2024-04-23
CV2-4528 Set the PERSISTENT_DISK_PATH for main alegre service.

sonoransun:
## Description
This change sets the PERSISTENT_DISK_PATH for the main alegre service.

Reference: CV2-4528

## How has this been tested?
A test deployment from this branch was made into QA, and the main alegre service confirmed to have this value set properly.

There are no security concerns with this change.


skyemeedan (approved):
I think the ticket number in PR title and description is for something else?

DGaffney (approved):

PR #395, opened 2024-04-25
CV2-4126 forcibly remove temporary media from consideration in search results

DGaffney:
## Description
Temporary results are squirming into search results - let's just not allow that. I don't think we will have a use case for them.

Reference: CV2-4126

## How has this been tested?
Tested on repl and confirmed to solve

NA


caiosba (approved):
Yes, I think we should never search for temporary object - on the contrary, temporary objects should be just search inputs.

PR #397, opened 2024-04-25
Force system wide python module installs if needed.

sonoransun:
## Description
While debugging GitLab CI issues, it was observed that the Alegre builds were failing when attempting to install the awscli (or other pip modules).

## How has this been tested?
A build and deploy from this branch confirmed the fix.

There are no security concerns with this change, as we previously performed system wide installations of pip modules and utiltiies.


skyemeedan (approved):
seems like a great work around! I wonder if there are base images we could use that already have the AWS utils installed?

PR #402, opened 2024-05-05
CV2-4126 Resolve broken video results

DGaffney:
## Description

Updates context querying logic and deletion logic, shuts off audio-video matching for now

Reference: CV2-4126

## How has this been tested?
Tested almost every permutation of similarity lookup locally and seems to work! Likely will require some test tweaks to accommodate.

None

### DIFF
--- a/app/main/controller/audio_similarity_controller.py
+++ b/app/main/controller/audio_similarity_controller.py
@@ -17,7 +17,7 @@ class AudioSimilarityResource(Resource):
     @api.doc('Delete an audio in the similarity database')
     @api.expect(audio_similarity_request, validate=True)
     def delete(self):
-        return similarity.delete_item(request.args, "audio")
+        return similarity.delete_item(request.json, "audio")
 
     @api.response(200, 'audio successfully stored in the similarity database.')
     @api.doc('Store an audio in the similarity database')
--- a/app/main/controller/video_similarity_controller.py
+++ b/app/main/controller/video_similarity_controller.py
@@ -17,7 +17,7 @@ class VideoSimilarityResource(Resource):
     @api.doc('Delete a video in the similarity database')
     @api.expect(video_similarity_request, validate=True)
     def delete(self):
-        return similarity.delete_item(request.args, "video")
+        return similarity.delete_item(request.json, "video")
 
     @api.response(200, 'video successfully stored in the similarity database.')
     @api.doc('Store a video in the similarity database')
--- a/app/main/lib/helpers.py
+++ b/app/main/lib/helpers.py
@@ -1,9 +1,26 @@
 def context_matches(query_context, item_context):
+  """
+    Check a pair of contexts to determine if they match - first pass is
+    typically overly permissive and based in postgres - this test does an
+    exact check. We do *not* check for equality or present for project_media_id,
+    as they are unique to all items, and as a result, if we did use it as a
+    comparator when looking at the contexts, we would never match any items, ever.
+    Add more globally unique context ids into that list. It's a hack for an
+    assumption we try to make everywhere else that we are agnostic about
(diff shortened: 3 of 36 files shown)


app/main/controller/audio_similarity_controller.py
```diff
@@ -17,7 +17,7 @@ class AudioSimilarityResource(Resource):
     @api.doc('Delete an audio in the similarity database')
     @api.expect(audio_similarity_request, validate=True)
     def delete(self):
-        return similarity.delete_item(request.args, "audio")
+        return similarity.delete_item(request.json, "audio") # Changed Since 4126 PR
```
computermacgyver, line 20: Are there any cases where we use `args` for delete in Check? I don't think so, but let's just double check. Same for Timpani @skyemeedan :pray:

app/main/controller/video_similarity_controller.py
```diff
@@ -17,7 +17,7 @@ class VideoSimilarityResource(Resource):
     @api.doc('Delete a video in the similarity database')
     @api.expect(video_similarity_request, validate=True)
     def delete(self):
-        return similarity.delete_item(request.args, "video")
+        return similarity.delete_item(request.json, "video") # Changed Since 4126 PR
```
computermacgyver, line 20: Same comment as above. Let's verify there are no DELETE requests with query-string arguments rather than bodies. I doubt there are, but worth checking.

app/main/lib/helpers.py
```diff
@@ -1,9 +1,26 @@
...
-      if item_context.get(k) != v and item_context.get(k) not in v:
-        return False
-    else:
-      if item_context.get(k) != v:
-        return False
-  return True
+    if k not in irrelevant_keys:
+      if isinstance(v, list):
+        if isinstance(item_context.get(k), list): # Changed Since 4126 PR
+          if not set(item_context.get(k)) & set(v): # Changed Since 4126 PR
```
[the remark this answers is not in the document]
computermacgyver: I'm hesitant to change things because it works, but I do agree with Code Climate here. Can we reverse the logic of some of these to avoid deeply nested `if`'s? E.g.,
```
    if k in irrelevant_keys:
      continue

    if not isinstance(v, list):
      if item_context.get(k) != v:
        return False
      else:
        continue
      
    if isinstance(item_context.get(k), list): # Changed Since 4126 PR
      if not set(item_context.get(k)) & set(v): # Changed Since 4126 PR
        return False # Changed Since 4126 PR
    else: # Changed Since 4126 PR
      if item_context.get(k) != v and item_context.get(k) not in v: # Changed Since 4126 PR
        return False # Changed Since 4126 PR
```

What is the purpose of `and item_context.get(k) not in v`?

app/main/lib/similarity_helpers.py
```diff
@@ -26,11 +26,11 @@ def drop_context_from_record(record, context):
     deleted = True
     return deleted
 
-def get_context_query(context, value_as_json=False, vars_as_hash=True):
+def get_context_query(context, value_as_json=True, vars_as_hash=True):
     context_query = []
     context_hash = {}
     for key, value in context.items():
-        if key not in ["project_media_id", "temporary_media"]:
+        if key not in ["project_media_id", "temporary_media", "content_type"]:
```
computermacgyver, line 33: It's not the specific changes made, but it seems like Line 38 is a bug here. 

```
                for i,v in enumerate(value):
                    if value_as_json:
                        context_clause += "context @> '[{\""+key+"\": "+json.dumps(value)+"}]'"
```

Shouldn't `value` be `v` in that line?

DGaffney: You're right on this one - I don't think this particular case comes up anywhere but good catch

computermacgyver (approved):

PR #411, opened 2024-05-13
CV2-4605 push up an initial idea to see how it plays against test suite

DGaffney:
## Description
Right now we dynamically load in `SentenceTransformers` any time we enqueue a task into alegre, and we do it needlessly - in order to drastically reduce our memory footprint and lessen errors associated with that bloat we should try to excise that importing from the web stack.

Reference: CV2-4605

## How has this been tested?
Not tested locally yet - establishing the basic changes needed first then coding up the rest.

None


app/main/lib/shared_models/shared_model.py
```diff
@@ -30,33 +30,46 @@ def import_model_class(model_class):
...
+  @staticmethod
+  def start_server(model_class, model_key, options={}):
+    app.logger.info('[%s] Serving model...', model_key)
+    SharedModel.register_server(model_class, model_key, options)
+    instance = SharedModel.get_server(model_key, options)
     instance.bulk_run()
 
+  @staticmethod
+  def get_model_info(model_key):
+    r = redis_client.get_client()
```
skyemeedan, line 58: if this is a static method, where is it getting the redis_client instance?

skyemeedan (commented):
I'm not clear on the design structure this is changing too.  Maybe one approach would be for the model_controller to keep track of the live models (vs looking them up per request)?  Like it could loop over all the model classes and inspect them (without loading to import) and keep in a dict?  Or since there is already "register" pattern, the registration of model keys could live with the controller instead of the SharedModel class?

app/main/lib/shared_models/shared_model.py
computermacgyver: Looks like `redis_client` is a global variable imported on line 12 `from app.main.lib import redis_client`

skyemeedan: oh!  does that mean we are instantiating new Redis clients with every request then?

DGaffney: @skyemeedan yes we are instantiating new Redis clients with every request - the latency for doing this is pretty minimal, and I've observed in other contexts before that holding them open for a while can actually induce problems. If you look at `check-api` we actually also use the same pattern of instantiating on use rather than having some abstracted session handler logic or a permanent open connection. I'm happy to change this if anyone wants in a separate ticket!

skyemeedan: clearly it has been working, but I'm wondering if that might be causing memory bloat?

DGaffney: I dont _think so_? After execution of this function, `r` would no longer be used and hence would be eligible for garbage collection. I don't think you _need_ to actually close a connection to redis either - or at least I haven't ever...

PR #414, opened 2024-05-17
CV2-4526 add flag to prevent bad needle reads

DGaffney:
## Description
Right now, there's a small amount of time where TMK files are not present on disk, but the video is stored in the database, and as such, queries could wrongfully search against missing TMK files (which raises the needle error). Flagging as finished will be a clean way to mitigate that going forwards. Will require deploy, and marking all previous videos as flagged after deploy.

Reference: CV2-4526

## How has this been tested?
Not yet tested locally - shaking down in tests here

None


skyemeedan (approved):
> and marking all previous videos as flagged after deploy.

Do we have a way to log in to db to do that? Maybe setting all pre-existing records to True could be part of the migration script?

PR #415, opened 2024-05-20
CV2-4600 add indices for slow querying on context

DGaffney:
## Description
We have a few slow query complaints coming from sentry - while our tables aren't too large, we can definitely improve queries with some indices on key fields and a catch-all on the entire `context` field.

Reference: CV2-4600

## How has this been tested?
built these indices locally and was able to get the migration to run successfully

None in particular!


computermacgyver (approved):
These all look like good indicies to have. I'm not sure that they are the source of the slow queries, which could be due to Alegre pushing its memory limits, but it makes sense to add these indicies regardless.

PR #418, opened 2024-05-24
CV2-4526 remove unnecessary haystack safety switch, add simple os path exists

DGaffney:
## Description
As mentioned in the Jira ticket, the error we run into isn't with the "haystack", but instead is with the "needle", the actual file being reviewed. I think the ultimate source of the problem is intermittent presto issues (practically, today, it is likely the timeouts we're seeing with tasks just because we haven't merged the timeout fix). There's nothing we can really *do* to save those files, so the move here is to essentially treat that file lookup as a 404 by pre-checking if the path exists. Also gutting the previous work as its more moving parts than we need to keep.

Reference: CV2-4526

## How has this been tested?
Not tested yet - should be straightforward as nothing in this is really new.

None


skyemeedan (commented):
In order to delete the migration, do we also need a rollback to remove the column?

DGaffney:
> In order to delete the migration, do we also need a rollback to remove the column?

I don't think anyone has actually pulled this into their local environment, nor has it gone out to any of our environments, so deleting it from the codebase should just reset us - @caiosba / @sonoransun does that sound right?

skyemeedan (approved):

caiosba:
@DGaffney you can confirm by check-in the deployed code, I guess.

PR #419, opened 2024-05-27
Cv2 4637 gut audio image

DGaffney:
## Description
Now that audio and image encoding are happening on presto, and there's no technical requirement on any packages when comparing rendered fingerprints for these items, we can release some of these dependencies from the codebase - this ticket does that!

Reference: CV2-4637

## How has this been tested?
Will need to review tests to make sure we're deleting safely but this should not have any new tests otherwise

None


skyemeedan (commented):
Whoohoo!  Everything will be so much lighter!   Do we need to remove any of the pdq, threatexchange build steps in the Dockerfile

DGaffney:
> Whoohoo! Everything will be so much lighter! Do we need to remove any of the pdq, threatexchange build steps in the Dockerfile

We can't yet because we depend on it for TMK afaik @computermacgyver please confirm? Mayyyybe we can get away with not building PDQ but that may not be the case.

skyemeedan (approved):

PR #421, opened 2024-05-28
CV2-4637 remove chromaprint from production dockerfile build

DGaffney:
## Description
This doesn't get run in develop - and I don't remember why we even have different dockerfiles - but we need to remove chromaprint from the prod file.

Reference: CV2-4637

## How has this been tested?
Not tested locally, for the above reason^

None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,7 @@ RUN apt-get update && apt-get install -y ffmpeg cmake swig libavcodec-dev libavf
 RUN ln -s /usr/bin/ffmpeg /usr/local/bin/ffmpeg
 
 # Copy necessary threatexchange folders
-COPY ./threatexchange/tmk/cpp /app/threatexchange/tmk/cpp
 COPY ./threatexchange/pdq/cpp /app/threatexchange/pdq/cpp
-RUN make -C /app/threatexchange/tmk/cpp
 
 # Other configurations
 RUN echo "set enable-bracketed-paste off" >> ~/.inputrc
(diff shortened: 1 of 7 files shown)


computermacgyver (approved):
I'm uncertain about us needing the 
```
RUN make -C /app/threatexchange/tmk/cpp
RUN make -C /app/threatexchange/tmk/cpp clean
```
lines as well. `tmkpy` should be taking care of all we need with tmk as this point. That said, I'm fine with this PR as is, and we can deal with TMK separately if needed.

DGaffney:
> I'm uncertain about us needing the
> 
> ```
> RUN make -C /app/threatexchange/tmk/cpp
> RUN make -C /app/threatexchange/tmk/cpp clean
> ```
> 
> lines as well. `tmkpy` should be taking care of all we need with tmk as this point. That said, I'm fine with this PR as is, and we can deal with TMK separately if needed.

I think we need TMK assets here because we have to run the needle analysis in alegre still - am I misunderstanding something?

computermacgyver:
> I think we need TMK assets here because we have to run the needle analysis in alegre still - am I misunderstanding something?

I'm uncertain. It really shouldn't be needed as [tmkpy](https://github.com/meedan/tmkpy) includes and compiles TMK as part of its installation (which happens via pip with it in `requirements.txt`). I don't think we should need TMK directly for anything at this point.

Dockerfile
```diff
@@ -7,9 +7,7 @@ RUN apt-get update && apt-get install -y ffmpeg cmake swig libavcodec-dev libavf
 RUN ln -s /usr/bin/ffmpeg /usr/local/bin/ffmpeg
 
 # Copy necessary threatexchange folders
-COPY ./threatexchange/tmk/cpp /app/threatexchange/tmk/cpp
 COPY ./threatexchange/pdq/cpp /app/threatexchange/pdq/cpp
```
computermacgyver, line 10: I think this PDQ line can go as well. It won't cause any problems if it stays, of course.

computermacgyver (approved):
Looks good. I think this is the right call to remove the TMK likes from the production Dockerfile. PDQ can also go from that file I believe.

PR #423, opened 2024-06-05
CV2-4699 add error logging for missing videos

DGaffney:
## Description
Adds direct logging for needle read failures where we attempt to read a file that doesn't yet exist

Reference: CV2-4699

## How has this been tested?
Not Tested locally yet but directly in line with how we log these types of inform errors other places.

None


computermacgyver (approved):
Looks good. Thanks!

PR #424, opened 2024-06-14
CV2-4752 remove pact from requirements (installed in dockerfiles already)

DGaffney:
## Description
For some reason we're getting issues with our Docker builds - the error is a SHA mismatch when installing pact-python - we already install it directly outside requirements.txt so going to see if just removing from the requirements.txt is sufficient

Reference: CV2-4752

## How has this been tested?
Not yet tested locally

None

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -80,7 +80,7 @@ oauthlib==3.1.0
 opensearch-py==2.2.0
 opt-einsum==3.2.0
 packaging==21.0
-pact-python==1.4.5
+# pact-python==1.4.5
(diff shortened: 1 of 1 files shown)


skyemeedan (approved):
Why not try it locally?

caiosba (approved):
I see the Travis build is passing now.

DGaffney:
> Why not try it locally?

I mean, I can, but the problem is in travis / CI builds, and it's just as easy to run it in CI than run it locally and heat up the laptop with some Docker...

PR #425, opened 2024-06-17
CV2-4719 add unique identifier as doc_id at alegre level

DGaffney:
## Description
Plumbs content_hash from check-api to presto request - not much exciting happening at this level

Reference: CV2-4719

## How has this been tested?
First changing this without updating tests to see what breaks, then fix tests, then integration tests locally

None relevant here

### DIFF
--- a/app/main/lib/presto.py
+++ b/app/main/lib/presto.py
@@ -26,6 +26,7 @@ class Presto:
     def send_request(presto_host, model_key, callback_url, message, requires_callback=True):
         data = {
             "callback_url": callback_url,
+            "content_hash": message.get("content_hash", str(uuid.uuid4())),
             "id": message.get("doc_id", str(uuid.uuid4())),
             "url": message.get("url"),
             "text": message.get("text"),
--- a/app/main/lib/presto.py
+++ b/app/main/lib/presto.py
@@ -26,13 +26,16 @@ class Presto:
     def send_request(presto_host, model_key, callback_url, message, requires_callback=True):
         data = {
             "callback_url": callback_url,
-            "content_hash": message.get("content_hash", str(uuid.uuid4())),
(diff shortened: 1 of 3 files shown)


app/main/lib/presto.py
```diff
@@ -26,6 +26,7 @@ def get_similar_items_callback_url(alegre_host, similarity_type):
     def send_request(presto_host, model_key, callback_url, message, requires_callback=True):
         data = {
             "callback_url": callback_url,
+            "content_hash": message.get("content_hash", str(uuid.uuid4())),
```
skyemeedan, line 29: I can see we definitely need a default random value if no hash provided so that it won't cache.  But in this case it isn't a hash.  I'm wondering if we need to know the type of hash used to know if we can compare hashes?  Or I guess non-comparable hashes will always result in non-matches so doesn't matter since the use case is caching?

DGaffney: Technically we don't need this, according to the spec on the presto side - given that I think I want to just remove the default value altogether!

app/main/lib/similarity.py
```diff
@@ -64,7 +64,7 @@ def get_body_for_text_document(params, mode):
       params['content'] = None
 
     if mode == 'store':
-      allow_list = set(['language', 'content', 'created_at', 'models', 'context', 'callback_url'])
+      allow_list = set(['language', 'content', 'created_at', 'models', 'context', 'callback_url', 'content_hash'])
```
skyemeedan, line 67: should doc_id be on the allow list?

DGaffney: I agree this is a bit wonky missing it here, but in all branches I'm looking at where its in store mode, the doc_id is handled separately.

skyemeedan (approved):

PR #428, opened 2024-07-03
CV2-4856 alter query body to check for vectors before querying against them

DGaffney:
## Description
Some queries on opensearch fail because we're querying on fields that don't exist on documents. This only happens on QA currently, but could happen anywhere because of some potential failure in storing documents correctly. We should be robust against those potential failures.

Reference: CV2-4856

## How has this been tested?
Works on QA servers when testing these changes directly

None

### DIFF
--- a/app/main/lib/text_similarity.py
+++ b/app/main/lib/text_similarity.py
@@ -108,23 +108,15 @@ def get_vector_model_base_conditions(search_params, model_key, threshold):
                     'bool': {
                         'must': [
                             {
-                                'match': {
-                                    'model_' + str(model_key): {
-                                        'query': "1",
-                                    }
+                                'exists': {
+                                    'field': 'vector_'+str(model_key)
                                 }
                             }
                         ]
                     }
                 },
                 'script': {
-                    'source': """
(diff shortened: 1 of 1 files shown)


app/main/lib/text_similarity.py
```diff
@@ -90,42 +90,49 @@ def get_elasticsearch_base_conditions(search_params, clause_count, threshold):
...
+                                        'query': "1",
+                                    }
+                                }
+                            }
+                        ]
+                    }
+                },
+                'script': {
+                    'source': """
+                        if (!doc.containsKey('vector_""" + str(model_key) + """') || doc['vector_""" + str(model_key) + """'].size() == 0) {
```
skyemeedan, line 122: Can you describe the intent of the logic change so we can check it?

computermacgyver:
I've pushed a slightly different approach as described on the JIRA ticket.

DGaffney:
@computermacgyver I've verified your version works equivalently on QA and am happy to merge if you or @skyemeedan signs off!

computermacgyver (approved):

PR #429, opened 2024-07-05
CV2-4790 fix broken error log due to missing file, surface s3 errors to sentry

DGaffney:
## Description
Right now we hit an error where `files` isn't defined when we try to log the needle error to sentry, and we don't log failures to download the files from S3 directly to sentry. Let's resolve that to make easier the job of surfacing whatever the needle error is.

Reference: CV2-4790

## How has this been tested?
No tests on this yet - may need to change a unit test or two.

None

### DIFF
--- a/app/main/lib/s3_client.py
+++ b/app/main/lib/s3_client.py
@@ -3,6 +3,8 @@ from botocore.client import Config
 
 from flask import current_app as app
 
+from app.main.lib.error_log import ErrorLog
+
 MINIO_HOST = "minio:9000"
 def download_file_from_s3(bucket: str, filename: str, local_path: str):
     """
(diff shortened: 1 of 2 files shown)


app/main/lib/s3_client.py
```diff
@@ -3,6 +3,8 @@
 
 from flask import current_app as app
 
+from app.main.lib.error_log import ErrorLog
```
skyemeedan, line 6: I'm surprised the ErrorLog is needed to get errors to sentry,  normally it already picks up on anything sent to python error output, so I'd think app.logger.error() would work?

DGaffney: Maybe its a difference in configuration but I believe Sentry only surfaces raised actual errors and not log messages?

skyemeedan: Maybe.  Originally I added a bunch of explicit sentry error dispatch, then removed most of it because I was seeing dupes. https://stackoverflow.com/questions/73456044/why-does-sentry-log-errors-caught-in-except-block

DGaffney: Hrm - well, I'll do the same then if I see the problem manifesting here? lol

skyemeedan (approved):

PR #432, opened 2024-07-10
CV2-4790 refactor check for file presence, add check for non-zero file

DGaffney:
## Description
In an ongoing attempt to catch the exact situation where this race condition happens, I'm doing two things here - first, move the file check to the exact moment before we need the file to try to mitigate any timing-based issues as much as possible, and second, add a constraint that the file also has to be actually filled with content, not just touch'ed or existing. Both of these have plausible theories for why they'd resolve this, potentially, but I don't personally think either are particularly likely. Again, it still looks like these are temporary items that are having these issues.

The theory for the temporary file is if we request an item be encoded twice (once for confirmed and another time for suggested or confirmed), and the item is a temporary item, then we could get into a situation where one of the workers deletes the file while the other worker is running the check. The theory for the file case is that perhaps the file is caught in some interstitial state (again unlikely, but maybe possible) and thus not actually ready for access.

If we see the issue less often with this patch my guess would be that we're likely running into the double-encoding case. A move for resolving it would be to potentially defer the deletion for a few minutes rather than deleting immediately?

Reference: CV2-4790

## How has this been tested?
Tested both tweaks against repl, should be fairly straightforward

None

### DIFF
--- a/app/main/lib/shared_models/video_model.py
+++ b/app/main/lib/shared_models/video_model.py
@@ -123,53 +123,54 @@ class VideoModel(SharedModel):
             result = Presto.blocked_response(response, "video")
             video.hash_value = result.get("body", {}).get("result", {}).get("hash_value")
         if video:
-            if self.tmk_file_exists(video):
-                matches = self.search_by_context(body["context"])
-                default_list = list(np.zeros(len(video.hash_value)))
-                try:
-                    l1_scores = np.ndarray.flatten((1-distance.cdist([r.get("hash_value", default_list) or default_list for r in matches], [video.hash_value], 'cosine'))).tolist()
-                except:
-                    app.logger.info('L1 scoring failed while running search for video id of '+str(video.id)+' match ids of : '+str([e.get("id") for e in matches]))
-                    l1_scores = [0.0 for e in matches]
-                qualified_matches = []
-                for i,match in enumerate(matches):
-                    if l1_scores[i] > app.config['VIDEO_MODEL_L1_SCORE']:
-                        qualified_matches.append(match)
-                files = self.get_fullpath_files(qualified_matches, False)
-                try:
+            matches = self.search_by_context(body["context"])
+            default_list = list(np.zeros(len(video.hash_value)))
+            try:
+                l1_scores = np.ndarray.flatten((1-distance.cdist([r.get("hash_value", default_list) or default_list for r in matches], [video.hash_value], 'cosine'))).tolist()
+            except:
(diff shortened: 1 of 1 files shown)


computermacgyver (changes requested):

app/main/lib/shared_models/video_model.py
```diff
@@ -155,21 +160,22 @@ def search(self, task):
...
                     return {"result": results[:limit]}
                 else:
                     return {"result": results}
             else:
-                ErrorLog.notify(Exception("Failed to locate needle for a video!"), {"video_folder": video.folder, "video_filepath": video.filepath, "video_id": video.id, "task": task})
                 return {"error": "Video not found for provided task", "task": task}
-        else:
-            return {"error": "Video not found for provided task", "task": task}
+        except Exception as err:
+            raise err
```
computermacgyver, line 171: Can we (do we need to) send to Sentry explicitly here? I think we should log in sentry if we're not certain that happens already

app/main/lib/shared_models/video_model.py
```diff
@@ -111,19 +111,20 @@ def search_by_context(self, context):
...
+            if video.hash_value is None:
+                callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], "video")
+                if task.get("doc_id") is    None:
+                    task["doc_id"] = str(uuid.uuid4())
+                response = json.loads(Presto.send_request(app.config['PRESTO_HOST'], "video__Model", callback_url, task, False).text)
+                # Warning: this is a blocking hold to wait until we get a response in 
+                # a redis key that we've received something from presto.
+                result = Presto.blocked_response(response, "video")
+                video.hash_value = result.get("body", {}).get("result", {}).get("hash_value")
+            if video:
```
computermacgyver, line 127: Technically we should be able to drop the `if video` conditional. We've already tried `video.hash_value` on line 118; so, if the video is None we'll have failed earlier than this line already. We must have a video at this point basically.

computermacgyver (approved):

app/main/lib/shared_models/video_model.py
DGaffney: Lets just add it for now - easy to see if we're double-logging and revert if necessary

app/main/lib/shared_models/video_model.py
DGaffney: agreed

app/main/lib/shared_models/video_model.py
DGaffney: Tweaked this commit so that we're not passing `video` attributes along since `video` wouldn't be guaranteed at this point of execution

PR #435, opened 2024-07-29
CV2-4790 add redownloading of tmk file to temp when missing

DGaffney:
## Description
Sometimes, because of race conditions, a TMK file is not available on disk when we are analyzing a match. This adds a patch to push in a temporary file during that time, and use that temporary file in lieu of the permanent one.

Reference: CV2-4790

## How has this been tested?
Tested line by line in REPL on server, should work fine

None

### DIFF
--- a/app/main/lib/shared_models/video_model.py
+++ b/app/main/lib/shared_models/video_model.py
@@ -110,19 +110,35 @@ class VideoModel(SharedModel):
             db.session.rollback()
             raise e
 
+    def get_blocked_response(self, task):
+        callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], "video")
+        if task.get("doc_id") is    None:
+            task["doc_id"] = str(uuid.uuid4())
+        response = json.loads(Presto.send_request(app.config['PRESTO_HOST'], "video__Model", callback_url, task, False).text)
+        # Warning: this is a blocking hold to wait until we get a response in 
+        # a redis key that we've received something from presto.
+        return Presto.blocked_response(response, "video")
+
+    def download_temp_file(self, task):
+        result = self.get_blocked_response(task).get("body")
+        s3_folder = (result.get("result", {}) or {}).get("folder")
+        s3_filepath = (result.get("result", {}) or {}).get("filepath")
+        tempfile = self.get_tempfile()
+        folder = str.join("/", tempfile.name.split("/")[0:-1])
+        filepath = tempfile.name.split("/")[-1]
+        with tempfile.NamedTemporaryFile(delete=False) as tmp:
+            folder = os.path.dirname(tmp.name)
+            filepath = os.path.basename(tmp.name)
+        download_file_from_s3(s3_folder, s3_filepath, media_crud.tmk_file_path(folder, filepath))
+        return folder, filepath
+
     def search(self, task):
         temporary = False
         try:
             body, threshold, limit = media_crud.parse_task_search(task)
             video, temporary = media_crud.get_object(body, Video)
             if video.hash_value is None:
-                callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], "video")
-                if task.get("doc_id") is    None:
-                    task["doc_id"] = str(uuid.uuid4())
-                response = json.loads(Presto.send_request(app.config['PRESTO_HOST'], "video__Model", callback_url, task, False).text)
-                # Warning: this is a blocking hold to wait until we get a response in 
-                # a redis key that we've received something from presto.
-                result = Presto.blocked_response(response, "video")
+                result = self.get_blocked_response(task)
                 video.hash_value = result.get("body", {}).get("result", {}).get("hash_value")
 
             matches = self.search_by_context(body["context"])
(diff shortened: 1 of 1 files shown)


caiosba (approved):
I can't tell if this is really going to fix the issue, but I see that code does what it's supposed to.

app/main/lib/shared_models/video_model.py
```diff
@@ -110,19 +110,35 @@ def search_by_context(self, context):
...
+        response = json.loads(Presto.send_request(app.config['PRESTO_HOST'], "video__Model", callback_url, task, False).text)
+        # Warning: this is a blocking hold to wait until we get a response in 
+        # a redis key that we've received something from presto.
+        return Presto.blocked_response(response, "video")
+
+    def download_temp_file(self, task):
+        result = self.get_blocked_response(task).get("body")
+        s3_folder = (result.get("result", {}) or {}).get("folder")
+        s3_filepath = (result.get("result", {}) or {}).get("filepath")
+        tempfile = self.get_tempfile()
```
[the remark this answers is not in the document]
caiosba: @DGaffney please confirm if this can be an issue or not... anyway, shouldn't we rename this variable just to be super safe with colisions?

DGaffney:
> I can't tell if this is really going to fix the issue, but I see that code does what it's supposed to.

I'm not quite sure yet myself, but this *should* solve for it. Temp file will force the file to exist, which will resolve the case. Since these happen only on temporary items, it shouldn't have some long-term effect either.

app/main/lib/shared_models/video_model.py
```diff
@@ -137,14 +153,18 @@ def search(self, task):
                 if l1_scores[i] > app.config['VIDEO_MODEL_L1_SCORE']:
                     qualified_matches.append(match)
             files = self.get_fullpath_files(qualified_matches, False)
+            if self.tmk_file_exists(video.folder, video.filepath):
+                folder, filepath = (video.folder, video.filepath)
+            else:
+                folder, filepath = self.download_temp_file(task)
```
computermacgyver, line 159: I'm concerned about this call @DGaffney . I think we should be passing in `result` and not `task` to this function. Unless there is a side-effect somewhere that I'm not seeing, `task` exists before we call Presto and does not have the location we actually want to download from.

Instead, what appears to be happening is that `download_temp_file` makes another call to `get_blocked_response`, which calls Presto another time. Can you check the logs locally and ensure Presto is only being called once for one innovation of this API endpoint on Alegre?

app/main/lib/shared_models/video_model.py
```diff
@@ -110,19 +110,35 @@ def search_by_context(self, context):
...
+        callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], "video")
+        if task.get("doc_id") is    None:
+            task["doc_id"] = str(uuid.uuid4())
+        response = json.loads(Presto.send_request(app.config['PRESTO_HOST'], "video__Model", callback_url, task, False).text)
+        # Warning: this is a blocking hold to wait until we get a response in 
+        # a redis key that we've received something from presto.
+        return Presto.blocked_response(response, "video")
+
+    def download_temp_file(self, task):
+        result = self.get_blocked_response(task).get("body")
```
computermacgyver, line 123: Per the comment below, I think `download_temp_file` should be having the `results` object from Presto as input and not be calling `get_blocked_response`, which calls Presto another time. At the time this function is called, we already have Presto output and should be using that rather than calling Presto a second time.

app/main/lib/shared_models/video_model.py
```diff
@@ -172,8 +192,8 @@ def search(self, task):
             if temporary:
                 self.delete(task)
```
computermacgyver, line 193: We should also try to explicitly delete `folder` and `tempfile` at this point if they exist (I don't see those updated values having been added to `task`

DGaffney: `folder` here is `tmp` and `tempfile` is a tempfile - I can certainly explicitly delete the tempfile but I believe it is, by definition, unnecessary?

app/main/lib/shared_models/video_model.py
DGaffney: I'm actually not sure about that @computermacgyver - the first call to Presto where `if video.hash_value is None:` is *never* true for the cases that we are treating - at the time of evaluation, the video exists, and it has a valid `hash_value`. Thus, we need a different check for if the file is missing. If you want, we could conflate both cases so they go through the same routine (e.g. `if video.hash_value is None or not self.tmk_file_exists(video.folder, video.filepath):`), but we'd still need to do the `if..else` block within that check, as we don't want to download a temporary version of the fingerprint in case its valid and already on disk:

```
if self.tmk_file_exists(video.folder, video.filepath):
    folder, filepath = (video.folder, video.filepath)
else:
    folder, filepath = self.download_temp_file(task)
```

If you want to ensure that we can only ever call `get_blocked_response` once for any possible iteration/combination of state, we can certainly refactor to accommodate that, but this particular issue is distinct from that existing `get_blocked_response` call, which handles a (iirc) much rarer edge case (and imo maybe even should be removed at this point)

PR #438, opened 2024-08-05
CV2-5011 refactors for making alegre dual purpose on text encoding

DGaffney:
## Description
In this ticket, I'm adding functionality for Alegre to serve text encoding responses both in the original, soon-to-be-deprecated way, as well as via presto. Idea is we'll serve both for a bit and slowly work out each individual case to use the new system over time.

Reference: CV2-5011

## How has this been tested?
Tested extensively locally, but I have not tested original pathways, so will be watching carefully with unit tests. Will also need to add more unit tests to accommodate all these changes.

None particularly relevant here.

### DIFF
--- a/app/main/lib/elastic_crud.py
+++ b/app/main/lib/elastic_crud.py
@@ -86,7 +86,7 @@ def parse_task_search(task):
     # alternatively, the "body" is just the entire dictionary.
     if "body" in task:
         body = task.get("body", {})
-        threshold = task.get("raw", {}).get('threshold', 0.0)
+        threshold = body.get("raw", {}).get('threshold', 0.0)
         limit = body.get("raw", {}).get("limit")
         if not body.get("raw"):
             body["raw"] = {}
--- a/app/main/lib/elastic_crud.py
+++ b/app/main/lib/elastic_crud.py
@@ -57,7 +57,7 @@ def get_blocked_presto_response(task, model, modality):
             if model_key != "elasticsearch" and not obj.get('model_'+model_key):
                 response = get_presto_request_response(model_key, callback_url, obj)
                 blocked_results.append(Presto.blocked_response(response, modality))
-        # Warning: this is a blocking hold to wait until we get a response in 
+        # Warning: this is a blocking hold to wait until we get a response in
         # a redis key that we've received something from presto.
         return obj, temporary, get_context_for_search(task), blocked_results[-1]
     else:
@@ -66,7 +66,7 @@ def get_blocked_presto_response(task, model, modality):
 def get_async_presto_response(task, model, modality):
     app.logger.error(f"get_async_presto_response: {task} {model} {modality}")
     obj, _ = get_object(task, model)
-    callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], modality)
+    callback_url = Presto.add_item_callback_url(app.config['ALEGRE_HOST'], modality)
     if task.get("doc_id") is None:
         task["doc_id"] = str(uuid.uuid4())
     task["final_task"] = "search"
--- a/app/main/lib/elastic_crud.py
+++ b/app/main/lib/elastic_crud.py
@@ -47,7 +47,7 @@ def requires_encoding(obj):
 def get_blocked_presto_response(task, model, modality):
     if task.get("doc_id") is None:
         task["doc_id"] = str(uuid.uuid4())
-    obj, _ = get_object(task, model)
+    obj, temporary = get_object(task, model)
     doc_id = obj["doc_id"]
     callback_url =  Presto.add_item_callback_url(app.config['ALEGRE_HOST'], modality)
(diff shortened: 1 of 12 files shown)


app/main/lib/elastic_crud.py
```diff
@@ -0,0 +1,99 @@
...
+    if context:
+        task["contexts"] = [context]
+    store_document(task, doc_id, language)
+    if task.get("content") and not task.get("text"):
+        task["text"] = task["content"]
+    return task, False
+
+def get_context_for_search(task):
+    context = {}
+    dup = copy.deepcopy(task)
```
skyemeedan, line 27: why do we make a copy here?

skyemeedan (commented):
I looked this over and it seemed fine, but I don't feel like I understand well enough to follow the details (including following the route where the old and new text pathways split).  If you need more detailed feedback, maybe we could pair and walk me through it?  

In addition to the unit tests,  did you try it out with local deployment and saw that when submitting a text query to alegre, it followed both the old and new pathways?

app/main/lib/elastic_crud.py
DGaffney: because if not, the pop mutates the task and affects it outside this function call, which is not what we want to do

app/main/lib/elastic_crud.py
```diff
@@ -0,0 +1,99 @@
...
+        task["doc_id"] = str(uuid.uuid4())
+    obj, temporary = get_object(task, model)
+    doc_id = obj["doc_id"]
+    callback_url = Presto.add_item_callback_url(app.config['ALEGRE_HOST'], modality)
+    app.logger.info(f"Object for {task} of model {model} with id of {doc_id} has requires_encoding value of {requires_encoding(obj)}")
+    if requires_encoding(obj):
+        blocked_results = []
+        for model_key in obj.pop("models", []):
+            if model_key != "elasticsearch" and not obj.get('model_'+model_key):
+                response = get_presto_request_response(model_key, callback_url, obj)
```
computermacgyver, line 58: We need to handle OpenAI embeddings at some place. Will these have a worker in Pesto or will Alegre call OpenAI directly?

app/main/lib/elastic_crud.py
```diff
@@ -0,0 +1,99 @@
...
+                blocked_results.append(Presto.blocked_response(response, modality))
+        # Warning: this is a blocking hold to wait until we get a response in
```
computermacgyver, line 60: Is there a timeout for Presto.blocked_response or can the code hang indefinitely. How do we proceed if there are multiple models and some but not all return vectors?

app/main/lib/text_similarity.py
```diff
@@ -53,8 +85,10 @@ def get_model_and_threshold(search_params):
       model_key = search_params['model']
   if 'threshold' in search_params:
       threshold = search_params['threshold']
-  if 'per_model_threshold' in search_params and search_params['per_model_threshold'].get(model_key):
+  if 'per_model_threshold' in search_params and isinstance(search_params['per_model_threshold'], dict) and search_params['per_model_threshold'].get(model_key):
       threshold = search_params['per_model_threshold'].get(model_key)
+  if 'per_model_threshold' in search_params and isinstance(search_params['per_model_threshold'], list) and [e for e in search_params['per_model_threshold'] if e["model"] == model_key]:
+      threshold = [e for e in search_params['per_model_threshold'] if e["model"] == model_key][0]["value"]
   if threshold is None:
       threshold = 0.9
```
computermacgyver, line 93: Let's please add a warning here that no threshold was supplied and 0.9 is being used as a default.

app/main/lib/elastic_crud.py
DGaffney: Good point. I see good arguments for both ways here - if you send that work to presto, you're centralizing concerns, not writing more code here, but adding some time to the work. If you do the opposite, vice versa. What do you think @computermacgyver?

app/main/lib/text_similarity.py
DGaffney: sounds good

app/main/lib/elastic_crud.py
DGaffney: Right now you'd wait indefinitely. We have a function `requires_encoding(obj)` that can determine whether a document has been fully analyzed - perhaps before we send out the response we throw a sentry event if we know it to only be partially fulfilled?

app/main/lib/elastic_crud.py
computermacgyver: I'd probably call OpenAI in Alegre. While it's a bit more code here, it is far less code than it would be on Presto's side to have a separate worker, queues, etc.

app/main/lib/elastic_crud.py
computermacgyver: I'm fine deferring this to future work, but I think we should have a timeout after which we just return whatever we have.

skyemeedan (approved):

PR #440, opened 2024-08-19
CV2-5082 add logic to suppress a callback event to check-api when explicitly requested

DGaffney:
## Description
In some cases we want to index items to alegre, but we don't care about completing a round trip request back to check-api. This adds an additional parameter that, in conjunction with https://github.com/meedan/check-api/pull/1996 will allow for the ability to suppress a response to Check API.

Reference: CV2-5082

## How has this been tested?
Not yet tested - will likely require its own unit test, for now lets see if it breaks anything

Nothing in particular of concern here!

### DIFF
--- a/app/main/controller/presto_controller.py
+++ b/app/main/controller/presto_controller.py
@@ -30,17 +30,20 @@ class PrestoResource(Resource):
         if action == "add_item":
             app.logger.info(f"Data looks like {data}")
             result = similarity.callback_add_item(data.get("body"), model_type)
-            if data.get("body", {}).get("raw", {}).get("final_task") == "search":
-                result = similarity.callback_search_item(data.get("body"), model_type)
-                if result:
-                    result["is_search_result_callback"] = True
-            callback_url = data.get("body", {}).get("raw", {}).get("callback_url", app.config['CHECK_API_HOST']) or app.config['CHECK_API_HOST']
-            if result and data.get("body", {}).get("raw", {}).get("requires_callback"):
-                app.logger.info(f"Sending callback to {callback_url} for {action} for model of {model_type} with body of {result}")
-                Webhook.return_webhook(callback_url, action, model_type, result)
-            output = {"action": action, "model_type": model_type, "data": result}
-            app.logger.info(f"PrestoResource value is {output}")
-            return_value = {"action": action, "model_type": model_type, "data": result}
+            if data.get("body", {}).get("raw", {}).get("suppress_response"):
+                return_value = {"action": action, "model_type": model_type, "data": result}
+            else:
+                if data.get("body", {}).get("raw", {}).get("final_task") == "search":
+                    result = similarity.callback_search_item(data.get("body"), model_type)
+                    if result:
+                        result["is_search_result_callback"] = True
+                callback_url = data.get("body", {}).get("raw", {}).get("callback_url", app.config['CHECK_API_HOST']) or app.config['CHECK_API_HOST']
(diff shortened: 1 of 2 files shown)


app/main/controller/presto_controller.py
```diff
@@ -30,17 +30,20 @@ def post(self, action, model_type):
...
-                    result["is_search_result_callback"] = True
-            callback_url = data.get("body", {}).get("raw", {}).get("callback_url", app.config['CHECK_API_HOST']) or app.config['CHECK_API_HOST']
-            if result and data.get("body", {}).get("raw", {}).get("requires_callback"):
-                app.logger.info(f"Sending callback to {callback_url} for {action} for model of {model_type} with body of {result}")
-                Webhook.return_webhook(callback_url, action, model_type, result)
-            output = {"action": action, "model_type": model_type, "data": result}
-            app.logger.info(f"PrestoResource value is {output}")
-            return_value = {"action": action, "model_type": model_type, "data": result}
+            if data.get("body", {}).get("raw", {}).get("suppress_response"):
+                return_value = {"action": action, "model_type": model_type, "data": result}
```
skyemeedan, line 34:
```suggestion
                # requested not to reply to caller with similarity response, so suppress it
                return_value = {"action": action, "model_type": model_type, "data": result}
```

app/main/controller/presto_controller.py
```diff
@@ -30,17 +30,20 @@ def post(self, action, model_type):
...
+                    result = similarity.callback_search_item(data.get("body"), model_type)
+                    if result:
+                        result["is_search_result_callback"] = True
+                callback_url = data.get("body", {}).get("raw", {}).get("callback_url", app.config['CHECK_API_HOST']) or app.config['CHECK_API_HOST']
+                if result and data.get("body", {}).get("raw", {}).get("requires_callback"):
+                    app.logger.info(f"Sending callback to {callback_url} for {action} for model of {model_type} with body of {result}")
+                    Webhook.return_webhook(callback_url, action, model_type, result)
+                output = {"action": action, "model_type": model_type, "data": result}
+                app.logger.info(f"PrestoResource value is {output}")
+                return_value = {"action": action, "model_type": model_type, "data": result}
```
skyemeedan, line 46: seems like this should return `output` since it is the same value? (and will be sure to align with what is logged)

app/main/controller/presto_controller.py
```diff
@@ -30,17 +30,20 @@ def post(self, action, model_type):
...
-            if result and data.get("body", {}).get("raw", {}).get("requires_callback"):
-                app.logger.info(f"Sending callback to {callback_url} for {action} for model of {model_type} with body of {result}")
-                Webhook.return_webhook(callback_url, action, model_type, result)
-            output = {"action": action, "model_type": model_type, "data": result}
-            app.logger.info(f"PrestoResource value is {output}")
-            return_value = {"action": action, "model_type": model_type, "data": result}
+            if data.get("body", {}).get("raw", {}).get("suppress_response"):
+                return_value = {"action": action, "model_type": model_type, "data": result}
+            else:
+                if data.get("body", {}).get("raw", {}).get("final_task") == "search":
```
skyemeedan, line 36:
```suggestion
                if data.get("body", {}).get("raw", {}).get("final_task") == "search":
                # compute a set of items that are similar to the just-stored item and respond to caller with them
```

skyemeedan (approved):
suggested some comments, and a tiny tweak

app/main/controller/presto_controller.py
DGaffney: fair enough!

PR #442, opened 2024-08-26

ahmednasserswe:
## Description


computermacgyver (approved):
Great to get this logging in place. Once tests pass, please merge.

PR #443, opened 2024-08-28

ahmednasserswe:

### DIFF
--- a/app/main/controller/image_ocr_controller.py
+++ b/app/main/controller/image_ocr_controller.py
@@ -3,7 +3,6 @@ from urllib3 import Retry
 from flask_restplus import Resource, Namespace, fields
 from google.cloud import vision
 import tenacity
-import json
 
 from app.main.lib.google_client import get_credentialed_google_client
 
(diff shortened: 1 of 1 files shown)


app/main/controller/image_ocr_controller.py
```diff
@@ -37,7 +38,7 @@ def post(self):
             return
```
skyemeedan, line 38: do we want to log "no text found in image using OCR" here so we can compare the number of misses?

app/main/controller/image_ocr_controller.py
```diff
@@ -37,7 +38,7 @@ def post(self):
...
 
         app.logger.info(
-            f"[Alegre OCR] [image_uri {image.source.image_uri}] Image OCR response package looks like {response}")
+            f"[Alegre OCR] [image_uri {image.source.image_uri}] Image OCR response package looks like {json.dumps(response)}")
```
skyemeedan, line 41: This seems like it is mostly for debugging, right, as it is logging the full response instead of just the text?

skyemeedan (approved):
Would it be possible/helpful to run this locally to be able to debug the structure of the response?

app/main/controller/image_ocr_controller.py
ahmednasserswe: we want to see the bounding_poly in the ocr response
gonna do a research task based on that. to determine what percentage of the image has to be text in order to be considered for ocr only.

PR #444, opened 2024-08-28

ahmednasserswe:


skyemeedan (approved):
Is there a reason we can't test locally to make sure that code executes to not spend time in PR loops?  (Or is it that we can't test responses from the OCR model locally because it only runs in QA?)

DGaffney (approved):

PR #448, opened 2024-09-05
CV2-2559: use local language detection model

computermacgyver:
## Description
Goal of this PR is to stop using Google's API for language identification and use a local model instead. We previously implemented CLD3, but had not turned it on. After some ad-hoc testing, I felt a FastText model was better. That model is included in this PR.

Longer term, language detection may belong in Presto. It's probably also not great to include the model in this repo, but in this case it's extremely small and including it here simplifies DevOps. Nonetheless, I welcome feedback on this point from the team.

Reference: CV2-2559

## How has this been tested?
Tested locally via flash shell. Existing tests cover the code and work.

N/A

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,6 +35,7 @@ deploy_qa:
     AWS_DEFAULT_REGION: $AWS_DEFAULT_REGION
   script:
     - apk add --no-cache curl python3 py3-pip git
+    - pip install setuptools==75.1.0
     - pip install urllib3==2.0.6
     - pip install botocore==1.31.62
     - pip install boto3==1.28.62
@@ -88,6 +89,7 @@ deploy_live:
     AWS_DEFAULT_REGION: $AWS_DEFAULT_REGION
   script:
     - apk add --no-cache curl jq python3 py3-pip git
+    - pip install setuptools==75.1.0
     - pip install urllib3==2.0.6
     - pip install botocore==1.31.62
     - pip install boto3==1.28.62


app/main/lib/elasticsearch.py
```diff
@@ -9,7 +9,8 @@
 from app.main.lib.language_analyzers import SUPPORTED_LANGUAGES
 from app.main.lib.error_log import ErrorLog
 #from app.main.lib.langid import Cld3LangidProvider as LangidProvider
-from app.main.lib.langid import GoogleLangidProvider as LangidProvider
+#from app.main.lib.langid import GoogleLangidProvider as LangidProvider
```
DGaffney, line 12: imo just delete

app/main/lib/langid.py
```diff
@@ -75,3 +76,24 @@ def langid(text):
   def test():
     cld3.get_language("Some text to check")
     return True
+
+class FastTextLangidProvider:
+# https://fasttext.cc/docs/en/language-identification.html
+  fasttext_model = fasttext.load_model("extra/fasttext_language_id/lid.176.ftz")
```
DGaffney, line 82: should we host this in s3 like (I believe) we do for other models?

app/main/lib/langid.py
```diff
@@ -75,3 +76,24 @@ def langid(text):
...
+  @staticmethod
+  def langid(text):
+    prediction = FastTextLangidProvider.fasttext_model.predict(text)
+    # prediction is a list of tupples, e.g., (('__label__en',), array([0.22517213]))
```
DGaffney, line 86: tuple typo

app/main/lib/text_similarity.py
```diff
@@ -7,7 +7,8 @@
 from app.main.lib.shared_models.shared_model import SharedModel
 from app.main.lib.language_analyzers import SUPPORTED_LANGUAGES
 #from app.main.lib.langid import Cld3LangidProvider as LangidProvider
-from app.main.lib.langid import GoogleLangidProvider as LangidProvider
+#from app.main.lib.langid import GoogleLangidProvider as LangidProvider
```
DGaffney, line 10: again I think just delete

DGaffney (approved):
some minor tweaks you can take or leave otherwise :shipit:

skyemeedan (commented):
I think the langid tests might need to be updated to explicitly check fasttext? https://github.com/meedan/alegre/blob/develop/app/test/test_langid.py  (some if it seems to test individual languages models independently)

computermacgyver:
> I think the langid tests might need to be updated to explicitly check fasttext? https://github.com/meedan/alegre/blob/develop/app/test/test_langid.py (some if it seems to test individual languages models independently)

Good point. I've added now :pray:

computermacgyver:
> Looks like one of the fasttext tests failed
> 
> ```
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/app/app/test/test_langid.py", line 83, in test_langid_fasttext
>     self.assertEqual(test['cld3'], result['result']['language'], test['text'])
> AssertionError: 'hi-Latn' != 'nl'
> - hi-Latn
> + nl
>  : namaste mera naam Karim hai
> ```
> 
> and there was another error about missing google credentials? (maybe no longer needed)

No this was me just running tests to see if fasttext would agree with the expected cld output in all cases. The answer is 'no' :joy: 

I think our language tests are not great because we have  a list of strings and their expected languages, but the expected language is different for different models :upside_down_face:

skyemeedan:
> I think our language tests are not great because we have a list of strings and their expected languages, but the expected language is different for different models 🙃

I'm guessing that is why the test was structured to not assume that results are identical across models, as long as they are consistent?  i.e. `hi` and `hi-latin` are both OK, but `nl` seems wrong

computermacgyver:
> I'm guessing that is why the test was structured to not assume that results are identical across models, as long as they are consistent? i.e. `hi` and `hi-latin` are both OK, but `nl` seems wrong

Sure, but our tests used to accept `en` from the Microsoft language id model. There are limitations with every model. CLD3 thinks, 'how to slice a banana' is `haw`, but fasttext identifies it correctly as `en`. There is no perfect model that will get every string right. Google's API does best, but we've been told to stop using. 

In this case, I believe fasttext hasn't been trained to detect languages in Latin characters that are not usually written with Latin characters. Notably it returns lower confidence scores for these examples.

I'm updating the tests to just skip these cases for fasttext now. I've left the other models alone for now. E.g., only Google gets this one right:
`{ 'fasttext': None, 'cld3': 'id', 'microsoft': 'fr', 'google': ['ta', 'ta-Latn'], 'text': 'vanakkam en peyar Karim' },`

I think this is a larger issue with the tests as written, but I'm not going to address that in this PR. The purpose of the tests as far as I'm concerned is to ensure the model produces consistent output across runs.

computermacgyver:
I've updated the code to fallback to Google when the fasttext result how low confidence. Taking this conversation internal to check if we have support for this.

app/main/lib/langid.py
computermacgyver: I'd like to not do so in this case as it would increase the complexity. With sentence transformers models, we automatically download and populate S3 with the model from hugging face if it doesn't exist, but fasttext doesn't have this fallback and it would have to be written manually, which seems overkill right now to avoid a 917KB file in the repo.

app/main/lib/langid.py
```diff
@@ -68,10 +71,71 @@ def langid(text):
...
+      'raw': prediction,
+      'model': 'FastText',
+    }
+
+  @staticmethod
+  def test():
+    FastTextLangidProvider.fasttext_model.get_language("Some text to check")
+    return True
+
+class HybridLangidProvider:
```
skyemeedan, line 112: Assuming the idea is we run both for a little while to see if the agreement is good enough (disagreement probably only on edge cases) and then disable CLD?

skyemeedan (approved):

app/main/lib/langid.py
```diff
@@ -68,10 +71,74 @@
...
+      # Now - FastText and CLD agree AND BOTH are more than 90% confident
+      # Return the higher confidence result
+      # if fasttext_result['result']['language'] != cld_result['result']['language']:
+      #   # Log when there is disagreement
+      #   app.logger.debug(json.dumps({
+      #     'service':'LangId',
+      #     'message': 'Disagreement between fasttext and cld. Returning higher confidence model',
+      #     'parameters':{'text':text, 'fasttext':fasttext_result, 'cld':cld_result,},
+      #     }))
+      if fasttext_result['result']['confidence'] > cld_result['result']['confidence']:
```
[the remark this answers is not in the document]
computermacgyver: I know, but I like it this way to be more explicit.

PR #450, opened 2024-09-26
CV2-5362 Set expiration explicitly on all SharedModel keys

DGaffney:
## Description
Extremely noncontroversial change - adds an expiry to SharedModel keys.

Reference: CV2-5362

## How has this been tested?
Not tested locally but very standard pattern

None

### DIFF
--- a/app/main/lib/shared_models/shared_model.py
+++ b/app/main/lib/shared_models/shared_model.py
@@ -85,6 +85,7 @@ class SharedModel(object):
 
   def send_response(self, task, response):
     self.datastore.set(task.task_id, json.dumps({"response": response}))
+    self.datastore.expire(task.task_id, 60*60*24)
 
   def run(self):
     while True:


skyemeedan:
seems like tests couldn't run because of build problem?

skyemeedan (approved):
I'm not familiar enough with the overall system to think through side effects of this. But setting the TTL seems like a great idea!  I'm assuming there are no situations we should be waiting longer than a day?  (indexing?). 

Was there a reason we suddenly started having lots more abandoned keys?  Do those failures show up anywhere else (sentry?) so we know it is happening?

computermacgyver (approved):

DGaffney:
> I'm not familiar enough with the overall system to think through side effects of this. But setting the TTL seems like a great idea! I'm assuming there are no situations we should be waiting longer than a day? (indexing?).
> 
> Was there a reason we suddenly started having lots more abandoned keys? Do those failures show up anywhere else (sentry?) so we know it is happening?

We don't really have any decent instrumentation on when these keys arrived or how long they've been there - technically, they could be there for as long as, god, four years? @computermacgyver @caiosba and I discussed during our weekly checkin and agreed that TTLing this for now, then checking to see if they keep coming up the next few days, was the solid strategy regardless.

PR #453, opened 2024-09-30
Hybrid local and remote lang id

computermacgyver:
## Description
Goal of this PR is to stop using Google's API for language identification and use a local model instead. We previously implemented CLD3, but had not turned it on. After some ad-hoc testing, I felt a FastText model was better. That model is included in this PR.

Longer term, language detection may belong in Presto. It's probably also not great to include the model in this repo, but in this case it's extremely small and including it here simplifies DevOps. Nonetheless, I welcome feedback on this point from the team.

Reference: CV2-2559

## How has this been tested?
Tested locally via flash shell. Existing tests cover the code and work.

N/A


DGaffney (approved):
Love this solution! Only concern I have is that we may be mixing and matching responses without knowing exactly which model was driving the result -is there any value in persisting that information for diagnostic purposes, somehow?

computermacgyver:
Thanks. Each result now includes a `model` key that indicates where the result came from. We also include more logging that I think should be sufficient for diagnostics.

PR #457, opened 2024-09-30
CV2-5181-Calculate-and-return-image_text_percentage-to-Check-API-as-part-of-the-OCR

ahmednasserswe:
## Description
calculating the area covered by text in an image

Reference: TICKET-ID CV2-5181

## How has this been tested?
tested by Unit tests.


app/main/controller/image_ocr_controller.py
```diff
@@ -37,9 +45,27 @@ def post(self):
         if not texts:
             return
```
skyemeedan, line 46: do we also want to log some thing here so we know how many images don't have text?  Or maybe we can already get that count from the post messages in logs?

skyemeedan (approved):

app/main/controller/image_ocr_controller.py
```diff
@@ -37,9 +45,27 @@ def post(self):
...
+        total_area = 0
+        for annotation in bounds:
+            vertices = [(v.x, v.y) for v in annotation.vertices]
+            area = self.polygon_area(vertices)
+            total_area += area
+        image_width = response.full_text_annotation.pages[0].width
+        image_height = response.full_text_annotation.pages[0].height
+        image_area = image_width * image_height
+        percentage = (total_area / image_area) * 100
+
```
computermacgyver, line 62: Let's put all the new code on a function and warp the call to that function in a try... except block

app/main/controller/image_ocr_controller.py
```diff
@@ -37,9 +45,27 @@ def post(self):
...
         app.logger.info(
-            f"[Alegre OCR] [image_uri {image.source.image_uri}] Image OCR response package looks like {convert_text_annotation_to_json(texts[0])}")
+            f"[Alegre OCR] [image_uri {image.source.image_uri}] [percentage of image area covered by text {percentage}%] Image OCR response package looks like {convert_text_annotation_to_json(texts[0])}")
 
+        # Assuming the image has a known width and height (you'll need to replace this with your actual image dimensions)
+        image_width = response.full_text_annotation.pages[0].width
+        image_height = response.full_text_annotation.pages[0].height
```
computermacgyver, line 68: Any reason for these three lines? They look line old debug code?

app/main/controller/image_ocr_controller.py
computermacgyver: Good point, but I think we're fine not to log as we can calculate that from other data as you suggest

app/main/controller/image_ocr_controller.py
ahmednasserswe: thanks for pointing that out

PR #460, opened 2024-10-10
Cv2 5370 media text split

DGaffney:
## Description
Small tweaks to make full end-to-end tiplines work locally

Reference: CV2-5370

## How has this been tested?
Tested extensively locally and confirmed to work

None


skyemeedan:
If you can give some more detail in the description about what the changes are intended to do, it would help me with reviewing more quickly.  Like it looks there is some simple naming refactoring, plus moving some calls to async blocking, adding paraphrase, and calling a separate endpoint for text based queries (I'm guessing this is the "media text split") in the title.  is that right?

app/main/lib/text_similarity.py
```diff
@@ -35,16 +35,29 @@ def get_document_body(body):
 def async_search_text(task, modality):
     return elastic_crud.get_async_presto_response(task, "text", modality)
 
+def sync_search_text(task, modality):
+    import code;code.interact(local=dict(globals(), **locals()))
```
skyemeedan, line 39: is code.interact a debugging line left in?

app/main/lib/text_similarity.py
```diff
@@ -175,6 +188,9 @@ def insert_model_into_response(hits, model_key):
             hit["_source"]["model"] = model_key
     return hits
 
+def return_sources(results):
+    return [dict(**r["_source"], **{"index": r["_index"], "score": r["_score"]}) for r in results]
```
skyemeedan, line 192: can you add a comment for what this does?  I think it is replacing the underscore version of the names?

skyemeedan (changes requested):

DGaffney:
> If you can give some more detail in the description about what the changes are intended to do, it would help me with reviewing more quickly. Like it looks there is some simple naming refactoring, plus moving some calls to async blocking, adding paraphrase, and calling a separate endpoint for text based queries (I'm guessing this is the "media text split") in the title. is that right?

That's basically right - I would summarize this as:

- Make all env_files consistent with mock s3 variables,
- Move the sync endpoint for text to use the blocking requests instead of legacy alegre-based fingerprinting,
- Making some minor refactors in `elastic_crud.py` so that we return the necessary data when running through pending fingerprinting tasks,
- Bypassing fields that we should never check against when querying objects (since they are either purely metadata or globally unique),
- Adding the paraphrase model,
- Structuring sync requests so that they can be properly immediately searched on,
- Clean up the response objects in `return_sources` to dis-embed the keys we need to send back to Check API,
- Test changes to accommodate all the above

Roughly in order of reading through the diff page.

app/main/lib/text_similarity.py
DGaffney: Yep Can do

app/main/lib/text_similarity.py
DGaffney: indeed

app/main/lib/similarity.py
```diff
@@ -200,6 +200,10 @@ def blocking_get_similar_items(item, similarity_type):
     response = video_model().blocking_search(model_response_package(item, "search"), "video")
     app.logger.info(f"[Alegre Similarity] [Item {item}, Similarity type: {similarity_type}] response for search was {response}")
     return response
+  elif similarity_type == "text":
+    response = sync_search_text(item, "text")
```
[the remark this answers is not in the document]
DGaffney: Nothing - I'd love to punt on cleaning up the convention until after we merge these changes and switch gears to refactoring and gutting legacy code if that is ok with you?

app/main/lib/elasticsearch.py
```diff
@@ -48,16 +48,17 @@ def generate_matches(context):
...
-            clause_count += len(context[key])
-            matches.append({
-                'query_string': { 'query': str.join(" OR ", [f"context.{key}: {v}" for v in context[key]])}
-            })
-        else:
-            clause_count += 1
-            matches.append({
-                'match': { 'context.' + key: context[key] }
-            })
+        if key not in ["project_media_id", "has_custom_id", "field"]:
```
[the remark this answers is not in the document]
DGaffney: Will add a little comment explicitly outlining intended logic!

.env_file.test
```diff
@@ -7,7 +7,7 @@ DATABASE_HOST=postgres
 DATABASE_USER=postgres
 DATABASE_PASS=postgres
 S3_ENDPOINT=http://minio:9000
-AWS_DEFAULT_REGION=eu-west-1
+AWS_DEFAULT_REGION=us-east-1
```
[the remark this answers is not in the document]
DGaffney: just inconsistent not "wrong" should have no actual effect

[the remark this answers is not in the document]
skyemeedan: Is there a reason not to have it align with everything else in eu-west-1?

app/main/lib/elasticsearch.py
```diff
@@ -45,19 +45,24 @@ def get_all_documents_matching_context(context):
     return []
 
 def generate_matches(context):
+    """
+        If the keys are not project_media_id, has_custom_id, or field, return ANDs for each field,
+        with ORs for intra-key values (e.g. foo = bar AND baz = (blah|bat))
```
skyemeedan, line 50: Aside: does this mean if I'm trying to restrict the search space by adding additional facets in the context, it will OR the result set?

skyemeedan (approved):

app/main/lib/elasticsearch.py
DGaffney: @skyemeedan it will OR values in a specific key, and AND across keys. So if you were trying to filter baz by blah *and* bat, it would not interpret that correctly and instead to blah|bat. Does that help at all?

skyemeedan: yeah I understand that, but just not clear what the reason for it.  It seems like, if I was trying to reduce the size of the result set to process (to avoid OpenSearch heap errors for example)  adding additional keys would expand the search space instead of reducing it.  There must be a feature that needs it to work this way?

PR #464, opened 2024-10-29
CV2-5050 remove some old logs from development work

DGaffney:
## Description
Very minor tweak to remove some erroneous logs that were either mislabeled or unnecessary outside development context.

Reference: CV2-5050

## How has this been tested?
Not Tested locally but expected to be.... obviously quite minimal.

None

### DIFF
--- a/app/main/lib/text_similarity.py
+++ b/app/main/lib/text_similarity.py
@@ -109,7 +109,7 @@ def get_model_and_threshold(search_params):
   if 'per_model_threshold' in search_params and isinstance(search_params['per_model_threshold'], list) and [e for e in search_params['per_model_threshold'] if e["model"] == model_key]:
       threshold = [e for e in search_params['per_model_threshold'] if e["model"] == model_key][0]["value"]
   if threshold is None:
-      app.logger.info(
+      app.logger.warn(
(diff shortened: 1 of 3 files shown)


app/main/lib/text_similarity.py
```diff
@@ -109,7 +109,7 @@ def get_model_and_threshold(search_params):
   if 'per_model_threshold' in search_params and isinstance(search_params['per_model_threshold'], list) and [e for e in search_params['per_model_threshold'] if e["model"] == model_key]:
       threshold = [e for e in search_params['per_model_threshold'] if e["model"] == model_key][0]["value"]
   if threshold is None:
-      app.logger.error(
+      app.logger.info(
```
computermacgyver, line 112: This might be best as `warn` . It's a situation we don't expect to see from Check and would be good to track and eliminate if it does happen.

computermacgyver (approved):

app/main/lib/text_similarity.py
caiosba: I agree with @computermacgyver , this should be a `warn` since it may be actionable.

caiosba (approved):

PR #465, opened 2024-10-30
CV2-5050 re-embed item on add event

DGaffney:
## Description
add_item needs to be embedded with an "item" key in order to play nicely with check-api
Reference: CV2-5050

## How has this been tested?
Relevant local tests appear to pass here, extremely minor tweaks if any are needed.

None

### DIFF
--- a/app/main/lib/similarity.py
+++ b/app/main/lib/similarity.py
@@ -133,7 +133,7 @@ def callback_add_item(item, similarity_type):
     if function:
         response = function(item)
         app.logger.info(f"[Alegre Similarity] CallbackAddItem: [Item {item}, Similarity type: {similarity_type}] Response looks like {response}")
-        return response
+        return {"item": response}
     else:
         app.logger.warning(f"[Alegre Similarity] InvalidCallbackAddItem: [Item {item}, Similarity type: {similarity_type}] No response")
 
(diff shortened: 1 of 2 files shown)


app/main/lib/similarity.py
```diff
@@ -133,7 +133,7 @@ def callback_add_item(item, similarity_type):
     if function:
         response = function(item)
         app.logger.info(f"[Alegre Similarity] CallbackAddItem: [Item {item}, Similarity type: {similarity_type}] Response looks like {response}")
-        return response
+        return {"item": response}
```
caiosba, line 136: Question from someone not familiarized enough with this code: I see that this condition applies to all types of media (audio, image, video and text)... just making sure that fixing it for text here this way won't break the other media types.

DGaffney: Yes - I checked with an image and does not appear to have an adverse effect - in fact, I don't believe there is *any* case where we run through this code-path with non-text at this time?

app/main/lib/similarity.py
```diff
@@ -133,7 +133,7 @@ def callback_add_item(item, similarity_type):
     if function:
         response = function(item)
         app.logger.info(f"[Alegre Similarity] CallbackAddItem: [Item {item}, Similarity type: {similarity_type}] Response looks like {response}")
```
dmou, line 135: drive-by comment: please also update the `Response looks like {response}` bit to reflect what's newly returned for more useful :wood:s thank u

computermacgyver (approved):

PR #468, opened 2024-11-04
CV2-5399 downgrade a bunch of info logs to debug, gut some logs, refactor others for consistency

DGaffney:
## Description
Cleaning up overly verbose logs - changes nothing about the execution of code itself!

Reference: CV2-5399

## How has this been tested?
Not tested locally, only change that may have an impact is missing references if new/shifted logs are in places where vars aren't, I don't think we have any of those though!

None

### DIFF
--- a/app/main/controller/audio_similarity_controller.py
+++ b/app/main/controller/audio_similarity_controller.py
@@ -33,5 +33,4 @@ class AudioSimilaritySearchResource(Resource):
     @api.doc(params={'url': 'audio URL to be stored or queried for similarity', 'threshold': 'minimum score to consider, between 0.0 and 1.0 (defaults to 0.9)', 'context': 'context'} )
     def post(self):
         args = request.json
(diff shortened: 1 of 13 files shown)


skyemeedan:
a lot of these seem like they really are debug logs (i.e. logging out full data of request) which will make logs really big and verbose in the long term.  For short term debugging, would it be easier to temporarily change the log level to DEBUG where it is deployed?

DGaffney:
> a lot of these seem like they really are debug logs (i.e. logging out full data of request) which will make logs really big and verbose in the long term. For short term debugging, would it be easier to temporarily change the log level to DEBUG where it is deployed?

Honestly at this point my preference is to just delete all these - @caiosba / @computermacgyver lmk what you think. @skyemeedan as it stands these are creating really big verbose logs _already_

ahmednasserswe (approved):

PR #470, opened 2024-11-07
CV2-5640 only send responses to check api when we have a fully realized response from presto during sync requests

DGaffney:
## Description
Updates to prevent cases where we send multiple partial matches back to check-api before we've run all vectorizations. The pop that we are removing is theorized as the culprit due to several factors:
1. The pop removes the model, which, on response, makes it appear that there are fewer items waiting to be processed,
2. The pop is only on sync requests, while the nearly identical async requests are using `get` and appear to be working correctly.
3. 
Reference: CV2-5640

## How has this been tested?
Not yet tested locally

None

### DIFF
--- a/app/main/lib/elastic_crud.py
+++ b/app/main/lib/elastic_crud.py
@@ -54,7 +54,7 @@ def get_blocked_presto_response(task, model, modality):
     callback_url = Presto.add_item_callback_url(app.config['ALEGRE_HOST'], modality)
     if requires_encoding(obj):
         blocked_results = []
-        for model_key in obj.pop("models", []):
+        for model_key in obj.get("models", []):
             if model_key != "elasticsearch" and not obj.get('model_'+model_key):
                 response = get_presto_request_response(model_key, callback_url, obj)
                 blocked_results.append({"model": model_key, "response": Presto.blocked_response(response, modality)})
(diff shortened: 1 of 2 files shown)


app/main/lib/elastic_crud.py
```diff
@@ -54,7 +54,7 @@ def get_blocked_presto_response(task, model, modality):
     callback_url = Presto.add_item_callback_url(app.config['ALEGRE_HOST'], modality)
     if requires_encoding(obj):
         blocked_results = []
-        for model_key in obj.pop("models", []):
+        for model_key in obj.get("models", []):
```
skyemeedan, line 57: Is the problem that we are destructively modifying the 'obj' object when the "models" list is removed?  It seems like it shouldn't make a difference since it isn't referenced later, but .get seems like better syntax.

skyemeedan (approved):
seems fine to me, but I don't really have full context.  Since this is kind of core functionality for presto I think we should have test coverage for it if it is breaking?

app/main/lib/elastic_crud.py
DGaffney: when we send these requests out, with the pop in place, it "tells" the alegre callback that is expecting fewer models than it really should be - by enforcing the get we (a) bring it in line with async and (b) probably eliminate erroneous calls back to check-api when alegre mistakenly things work is complete for every in-process vector

PR #474, opened 2024-12-19
CV2-5778-change-language-detection-agreement

ahmednasserswe:
…TH of them is more than "Threshold" confident`

lowering threshold to 0.7

## Description
implementing this strategy to use google language detection api 
Both CLD3 and FastText must agree and BOTH of them is more than "Threshold" confident
Threshold here is 0.7 
Practically, in this PR, we only lower the threshold from 0.9 to 0.7

Reference: CV2-5367 (to provide additional context)

## How has this been tested?
unittests

### DIFF
--- a/app/main/lib/langid.py
+++ b/app/main/lib/langid.py
@@ -117,9 +117,10 @@ class HybridLangidProvider:
     fasttext_result = FastTextLangidProvider.langid(text)
     cld_result = Cld3LangidProvider.langid(text)
     # max_confidence = max(fasttext_result['result']['confidence'], cld_result['result']['confidence'])
+    min_confidence = min(fasttext_result['result']['confidence'], cld_result['result']['confidence'])
 
     # if fasttext_result['result']['language'] == cld_result['result']['language'] or max_confidence >= 0.8:
-    if fasttext_result['result']['language'] is not None and cld_result['result']['language']  is not None and fasttext_result['result']['language'] == cld_result['result']['language'] and min(fasttext_result['result']['confidence'], cld_result['result']['confidence']) >= 0.7:
+    if fasttext_result['result']['language'] == cld_result['result']['language'] and min_confidence >= 0.7:
       # OLD - FastText and CLD agree or one of them is more than 80% confident.
       # Now - FastText and CLD agree AND BOTH are more than 90% confident
       # Return the higher confidence result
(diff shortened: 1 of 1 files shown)


computermacgyver (changes requested):
Can you please address the Sentry error that has been tagged on this PR?

It occurs when one of the confidence scores is `None`. We can avoid this by removing the `min_confidence=` line and putting this directly in the `if` statement alongside a check that one of the values is not `None`

computermacgyver (approved):

app/main/lib/langid.py
```diff
@@ -117,10 +117,12 @@ def langid(text):
     fasttext_result = FastTextLangidProvider.langid(text)
     cld_result = Cld3LangidProvider.langid(text)
     # max_confidence = max(fasttext_result['result']['confidence'], cld_result['result']['confidence'])
```
skyemeedan, line 119: please remove the disabled versions of the code, unless there is a reason we need to keep it around, in which case please add a comment to explain why

app/main/lib/langid.py
```diff
@@ -117,10 +117,12 @@ def langid(text):
...
-    min_confidence = min(fasttext_result['result']['confidence'], cld_result['result']['confidence'])
+    # min_confidence = min(fasttext_result['result']['confidence'], cld_result['result']['confidence'])
 
     # if fasttext_result['result']['language'] == cld_result['result']['language'] or max_confidence >= 0.8:
-    if fasttext_result['result']['language'] == cld_result['result']['language'] and min_confidence >= 0.9:
+    if fasttext_result['result']['confidence'] is not None and cld_result['result']['confidence'] is not None \
+            and fasttext_result['result']['language'] == cld_result['result']['language'] \
+            and  min(fasttext_result['result']['confidence'], cld_result['result']['confidence']) >= 0.7:
```
skyemeedan, line 125: Can you add a commented with the currently expected behavior `Both CLD3 and FastText must be have non-null confidence scores, agree on language tag, and BOTH of them is more than "Threshold" confident` and the link to the artifact where the decision was made about the 0.7 threshold value?  I think that is: https://meedan.atlassian.net/browse/CV2-5367

app/main/lib/langid.py
```diff
@@ -117,10 +117,12 @@ def langid(text):
...
       # OLD - FastText and CLD agree or one of them is more than 80% confident.
       # Now - FastText and CLD agree AND BOTH are more than 90% confident
```
skyemeedan, line 127: I think these comments not needed anymore?

skyemeedan (commented):
Logic looks great!  can you please clean up the comments a bit so it will not be confusing next time we look at this code?

skyemeedan (approved):

PR #476, opened 2025-01-02
CV2-5890 remove any ties to shared model for audio model

DGaffney:
## Description
Guts any audio relationships to shared model - looks like chromaprint / "heavy" dependencies are gone, but this fully separates so that we can safely turn off audio workers.

Reference: CV2-5890

## How has this been tested?
Not yet tested locally, watching tests in CI

None


skyemeedan:
* Are we able to remove any of the psql audio comparison functions from Alegre? do they belong to the Presto audio model now? https://github.com/meedan/alegre/blob/develop/manage.py
* If we removed the chromaprint dependencies in May,  how has it been working all this time? https://github.com/meedan/alegre/commit/090546e2ac613868b77992dfe5aa45314615d782

DGaffney:
> * Are we able to remove any of the psql audio comparison functions from Alegre? do they belong to the Presto audio model now? https://github.com/meedan/alegre/blob/develop/manage.py
> * If we removed the chromaprint dependencies in May,  how has it been working all this time? [090546e](https://github.com/meedan/alegre/commit/090546e2ac613868b77992dfe5aa45314615d782) ... that must be when we deployed the audio and image presto workers?

We cannot remove the psql audio comparison functions, because that is the mechanism by which we do similarity lookups. That has to stick around, unfortunately. And yes - we moved all this to presto in may, which is why it works.

skyemeedan (approved):
Looks great to me!

PR #479, opened 2025-01-10
[CV2-5845] error when text for similarity is missing

skyemeedan:
## Description

Adding assertion so that attempting to process a similarity request with content or text missing or None will cause error (and assuming exception will result in error code returned by flask)

This caused some of the tests of the bulk similarity endpoint to fail because they didn't include text in the test payload, so I added some dummy text

Also realized that the similarity tests were still executing against the old text similarity endpoint (and I couldn't find any corresponding tests for the new endpoint) so switched them to the new endpoint

Reference: https://meedan.atlassian.net/browse/CV2-5874

## How has this been tested?
not yet

Please list any security concerns that may be relevant.

### DIFF
--- a/app/main/lib/similarity.py
+++ b/app/main/lib/similarity.py
@@ -53,6 +53,9 @@ def get_body_for_text_document(params, mode):
       params['content'] = params.get('text')
       del params["text"]
 
+    # Confirm that content is not empty or missing
+    assert (params['content'] is not None and params['content'] != ""), "Similarity cannont be computed for null or missing content text"
+
     # Set defaults
     if 'created_at' not in params:
       params['created_at'] = datetime.now()
--- a/app/main/lib/similarity.py
+++ b/app/main/lib/similarity.py
@@ -54,7 +54,7 @@ def get_body_for_text_document(params, mode):
       del params["text"]
 
     # Confirm that content is not empty or missing
-    assert (params['content'] is not None and params['content'] != ""), "Similarity cannot be computed for null or missing content text"
+    assert (params.get('content') is not None and params['content'] != ""), "Similarity cannot be computed for null or missing content text"
 
     # Set defaults
     if 'created_at' not in params:
--- a/app/main/lib/similarity.py
+++ b/app/main/lib/similarity.py
@@ -53,8 +53,8 @@ def get_body_for_text_document(params, mode):
       params['content'] = params.get('text')
       del params["text"]
 
-    # Confirm that content is not empty or missing
-    assert (params.get('content') is not None and params['content'] != ""), "Similarity cannot be computed for null or missing content text"
+    # Confirm that content is not empty, missing, or whitespace
+    assert (params.get('content') is not None and params['content'].strip() != ""), "Similarity cannot be computed for null or missing content text"
 
     # Set defaults
     if 'created_at' not in params:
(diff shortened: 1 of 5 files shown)


app/test/test_similarity_lang_analyzers.py
```diff
@@ -33,7 +33,7 @@ def test_all_analyzers(self):
         examples = [{ 'text': 'केले को कैसे काटें', 'language': 'hi'}, {'text': 'how to slice a banana', 'language': 'en'}, {'text': 'como rebanar un plátano', 'language': 'es'}, {'text': 'কিভাবে একটি কলা টুকরা করা হয়', 'language': 'bn'}]
         with self.client:
             for example in examples:
-                response = self.client.post('/text/similarity/', data=json.dumps(example), content_type='application/json')
+                response = self.client.post('/similarity/sync/text/', data=json.dumps(example), content_type='application/json')
```
DGaffney, line 36: I think you probably need to add more structure to the `example` here - see something like `test_audio_basic_http_responses` in `test_sync_similarity.py` (and probably mock the response from Presto as we do there as well). Should just be a matter of adding a few more keys here:

```
{
    'text': 'केले को कैसे काटें',
    'language': 'hi',
    'project_media_id': 1,
    'context': {
        'team_id': 1,
    }
}
```

app/main/lib/similarity.py
```diff
@@ -53,6 +53,9 @@ def get_body_for_text_document(params, mode):
       params['content'] = params.get('text')
       del params["text"]
 
+    # Confirm that content is not empty or missing
+    assert (params.get('content') is not None and params['content'] != ""), "Similarity cannot be computed for null or missing content text"
```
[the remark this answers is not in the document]
computermacgyver: Good point. I agree that we should reject whitespace only strings. We can do this easily by changing `params['content'] != "")` to `params['content'] .strip() != "") `

app/test/test_similarity_lang_analyzers.py
computermacgyver: We shouldn't have to mock presto here. We can avoid calls to Presto all together by leaving `models` off (as is) and/or setting it equal to `['elasticsearch']` so that we're not using any vector models and hence have no Presto calls.

app/main/lib/similarity.py
```diff
@@ -53,6 +53,9 @@ def get_body_for_text_document(params, mode):
       params['content'] = params.get('text')
       del params["text"]
 
+    # Confirm that content is not empty, missing, or whitespace
+    assert (params.get('content') is not None and params['content'].strip() != ""), "Similarity cannot be computed for null or missing content text"
```
skyemeedan, line 57:
```
======================================================================
FAIL: test_model_similarity_without_text (test_similarity.TestSimilarityBlueprint)
This should return an error because model cannot compute on empty text
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/app/test/test_similarity.py", line 547, in test_model_similarity_without_text
    assert response.status_code >= 500, f"response status code was {response.status_code}"
AssertionError: response status code was 404
```
@DGaffney, the new endpoint returns 404 for an assertion error instead of 500, is this by design? do you know how I should update it so the framework will 402 or 500 error?

DGaffney (approved):
:shipit: !

PR #481, opened 2025-01-13
CV2-5992 fix bad variable name

DGaffney:
## Description
Minor tweak, no idea how this would have ever worked without the tweak but we need to not step on an imported variable for a variable name.

Reference: CV2-5992

## How has this been tested?
Unit test covers this function run such that a similarly bad named case would yield an error in the future

None

### DIFF
--- a/app/main/lib/shared_models/video_model.py
+++ b/app/main/lib/shared_models/video_model.py
@@ -121,9 +121,9 @@ class VideoModel():
         result = self.get_blocked_response(task).get("body")
         s3_folder = (result.get("result", {}) or {}).get("folder")
         s3_filepath = (result.get("result", {}) or {}).get("filepath")
-        tempfile = self.get_tempfile()
-        folder = str.join("/", tempfile.name.split("/")[0:-1])
-        filepath = tempfile.name.split("/")[-1]
+        temp_file = self.get_tempfile()
+        folder = str.join("/", temp_file.name.split("/")[0:-1])
+        filepath = temp_file.name.split("/")[-1]
         with tempfile.NamedTemporaryFile(delete=False) as tmp:
             folder = os.path.dirname(tmp.name)
             filepath = os.path.basename(tmp.name)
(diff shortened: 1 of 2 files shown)


DGaffney:
> * It says it was not tested locally. Since we know that the tests were not catching this error on the expected execution path before, can you confirm that the mocked test generates a similar error to the one observed in sentry?

The unit test was added after that note to test it - I will update the main note accordingly. It is now covered in a test.

skyemeedan:
I was asking if the unit test generates a similar error to the one in sentry because, It doesn't seem to me that the unit test covers the error behavior.  Like with or without the fix it returns 'path' as the path, vs triggering `'_io.BufferedRandom' object has no attribute 'NamedTemporaryFile'`  Did you see the error when testing by hand?

DGaffney:
> I was asking if the unit test generates a similar error to the one in sentry because, It doesn't seem to me that the unit test covers the error behavior. Like with or without the fix it returns 'path' as the path, vs triggering `'_io.BufferedRandom' object has no attribute 'NamedTemporaryFile'` Did you see the error when testing by hand?

Ah I see - the mock actually hides the issue - test now updated to accommodate an actual live fire, and I've walked through it manually to check that it is in fact the right type of object that would have triggered the original issue in production.

DGaffney:
@skyemeedan apologies - I had pushed the fix yesterday to a different branch that had been worked off of master since I was last in alegre during a deploy - didn't realize it was on the wrong branch - the fix is on here now!

skyemeedan (approved):
Thanks for fixing the test!  Now running with the old code reproduces the error nicely
```
ERROR: test_download_temp_file (test_video_similarity.TestVideoSimilarityBlueprint)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/app/test/test_video_similarity.py", line 147, in test_download_temp_file
    folder, filepath = self.model.download_temp_file(task)
  File "/app/app/main/lib/shared_models/video_model.py", line 127, in download_temp_file
    with tempfile.NamedTemporaryFile(delete=False) as tmp:
  File "/usr/local/lib/python3.7/tempfile.py", line 476, in __getattr__
    a = getattr(file, name)
AttributeError: '_io.BufferedRandom' object has no attribute 'NamedTemporaryFile'
```

PR #484, opened 2025-01-28
update CODEOWNERS to include infrastructure team

skyemeedan:
update CODEOWNERS to include infrastructure team for changes involving deloyment and containers

## Description
As discussed on to align across services and policies

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,4 @@
 * @DGaffney @skyemeedan @computermacgyver
+/.github/ @dmou
+/production/ @dmou
+Dockerfile @dmou
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,3 +2,4 @@
 /.github/ @dmou
 /production/ @dmou
 Dockerfile @dmou
+docker* @dmou


caiosba (approved):

caiosba (changes requested):
Please add the `docker-*` files as well (like Check API and Pender).

skyemeedan:
> Please add the `docker-*` files as well (like Check API and Pender).

Pender has `/docker* @dmou` which I thought was matching docker directories

caiosba:
It's matching docker-compose.yml, docker-test.yml, etc.

skyemeedan:
> It's matching docker-compose.yml, docker-test.yml, etc.

oh I see, the slash restricts to only matching in toplevel dir
https://git-scm.com/docs/gitignore#_pattern_format
> if there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

DGaffney (approved):
:shipit:

PR #485, opened 2025-02-05
[CV2-6020] test updates and handling for missing min_es_score

skyemeedan:
## Description

adds more test around min_es_score behavior, updates to 

> For updating the behavior, seem the goal would be
min_es_score missing or None: set to zero  with warning
min_es_score cannot parse as float: log error and raise exception?

Reference: https://meedan.atlassian.net/browse/CV2-6020

## How has this been tested?
Test coverage added, ran locally,
* confirmed warning logged appropriately from test `WARNING:flask.app:min_es_score is missing or None, defaulting to 0.0`
* confirmed Exception raised by float parsing is logged appropriately `ERROR:flask.app:Unable to parse min_es_score 'fifty':could not convert string to float: 'fifty'`
* assertion checks that error case results in 500
* 
no new security concerns

### DIFF
--- a/app/main/lib/text_similarity.py
+++ b/app/main/lib/text_similarity.py
@@ -201,13 +201,20 @@ def strip_vectors(results):
     return results
 
 def restrict_results(results, search_params, model_key):
+    """
+    min_es_score is the minimum elasticsearch score needed to include a result. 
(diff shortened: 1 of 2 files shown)


DGaffney (approved):
Was running into the same problems on CV2-6022 - this looks basically correct!

PR #487, opened 2025-02-06
CV2-6022 update tests to use non-deprecated URLs

DGaffney:
## Description
Updates tests to move to sync endpoint in lieu of soon to be deprecated endpoint

Reference: CV2-6022

## How has this been tested?
Ran through most of these test updates individually, one at a time. Should be good to go!

None


app/test/data/similarity.json
```diff
@@ -1,9 +1,11 @@
 [
   {
-    "content": "This is a test"
+    "content": "This is a test",
+    "context": {"dbid": 123}
```
skyemeedan, line 4: It looks like many (all?) the examples now have a dbid in the context, is this a required element now for the calls to the new endpoints?   If so, are there upstream services that will need to add this data element to their text similarity calls?

app/test/test_similarity.py
```diff
@@ -407,202 +440,56 @@ def test_elasticsearch_delete_text(self):
...
-              'context': {
-                'dbid': 54
-              }
-            }),
-            content_type='application/json'
-        )
-        result = json.loads(response.data.decode())
-        self.assertEqual(0, len(result['result']))
-
-    def test_model_similarity_without_text(self):
```
skyemeedan, line 540: This definitely shouldn't be deleted.  It uses the new endpoint and elastic search model.  (was added yesterday in the PR)

app/test/test_similarity.py
```diff
@@ -407,202 +440,56 @@ def test_elasticsearch_delete_text(self):
...
+                  'language': 'hi',
+                  'context': { 'dbid': 543 },
+                  'min_es_score': 0.1,
                 }),
                 content_type='application/json'
             )
             result = json.loads(response.data.decode())
             self.assertEqual(1, len(result['result']))
 
-    def test_model_similarity(self):
```
skyemeedan, line 442: Instead of deleting these tests, lets move them to a separate file like `similarity_tests_deprecated.py`. Then we can hopefully migrate them once we move the CI testing to an environment that has access to presto models for vectorization.

app/test/test_sync_similarity.py
```diff
@@ -22,6 +25,96 @@ def tearDown(self): # done in our pytest fixture after yield
         db.session.remove()
         db.drop_all()
 
+    def test_text_basic_http_responses_with_doc_id(self):
+        text = 'This is some sample text to test with'
```
skyemeedan, line 29: Can you add a comment here about which component of Alegre is being tested? (i.e. what is the client object?)  It seems like we are basically just asserting the presence of the same data elements we included in the mocK?

app/test/test_text_similarity.py
```diff
@@ -1,60 +0,0 @@
-import unittest
```
skyemeedan, line 1: It doesn't seem like these tests should be deleted?  It seems like they are validating the setup of the document structure to be inserted, and don't involve actual calls to presto?

app/test/data/similarity.json
DGaffney: `dbid` is not required, but having _any `context` at all_ is required - and I do believe all upstream services already do this.

app/test/test_similarity.py
DGaffney: Will revert that part

app/test/test_similarity.py
DGaffney: Will move to a deprecation file

app/test/test_text_similarity.py
DGaffney: That's fair I suppose!

skyemeedan:
@DGaffney I'm not clear why so many tests have been moved to the `app/old/` directory.  Are we no longer supporting features like OCR or audio transcription?

DGaffney:
@skyemeedan should be good to go here now!

app/test/test_similarity_deprecated.py
```diff
@@ -0,0 +1,176 @@
...
+              'context': {
+                'dbid': 54
+              }
+            }),
+            content_type='application/json'
+        )
+        result = json.loads(response.data.decode())
+        self.assertEqual(0, len(result['result']))
+
+    def test_model_similarity_without_text(self):
```
skyemeedan, line 125: `test_model_similarity_without_text` isn't deleted, should stay in original file? (and remove from here)

app/test/test_similarity_deprecated.py
```diff
@@ -0,0 +1,176 @@
...
+from opensearchpy import helpers, OpenSearch, TransportError
+from flask import current_app as app
+import numpy as np
+
+from app.main import db
+from app.test.base import BaseTestCase
+from app.main.lib.shared_models.shared_model import SharedModel
+from unittest.mock import patch
+
+class TestSimilarityBlueprint(BaseTestCase):
```
skyemeedan, line 12:
```suggestion
class TestSimilarityBlueprintDeprecated(BaseTestCase):
"""
These tests involve text vectorization functionality provided by presto.  
# TODO: update to new similarity endpoints after vectorization is available in CI env: https://meedan.atlassian.net/browse/CV2-6160
"""

```

app/test/test_similarity_lang_analyzers.py
```diff
@@ -90,41 +89,44 @@ def test_auto_language_id(self):
...
               result = json.loads(response.data.decode())
               # indirectly checking classification by confirming which index was included in result
               index_alias = app.config['ELASTICSEARCH_SIMILARITY']
-              if expected_lang is not None:
-                  index_alias = app.config['ELASTICSEARCH_SIMILARITY']+"_"+expected_lang
-              self.assertTrue(index_alias in [e['index'] for e in result['result']])
+              index_alias_language = app.config['ELASTICSEARCH_SIMILARITY']+f"_{expected_lang}"
+              indices = [e['index'] for e in result['result']]
+              self.assertTrue(
+                  index_alias in indices or index_alias_language in indices,
```
skyemeedan, line 127: won't this now return true when it is stored in the default ES index only? i.e. with the 'or' it is no longer validating that 'auto' guessed the correct language and it was added to the correct language index?   should be 'and'?

app/test/test_similarity.py
```diff
@@ -613,36 +500,23 @@ def test_min_es_search(self):
...
             response = self.client.post(
-                '/text/similarity/search/',
+                '/similarity/sync/text',
                 data=json.dumps(data),
                 content_type='application/json'
             )
             result = json.loads(response.data.decode())
             self.assertEqual(0, len(result['result']))
 
-            # confirm that min_es_score missing or None: set to zero  with warning
```
skyemeedan, line 623: why are these tests removed?

skyemeedan (changes requested):

app/test/test_similarity.py
DGaffney: weird that its not showing up even though its in branch - will try something to put it back in place

app/test/test_similarity_lang_analyzers.py
DGaffney: Possibly? I'll revert and we'll see what it does.

skyemeedan:
>  this looks to be failing just on the language test, but at first blush, I don't even know if this is really a failure mode? It would seem to me that this could be an entirely intended end-state of this test, no?

@DGaffney The intention of the test is to validate the behavior that text of one of the recognized languages is provided, it is added to the appropriate language specific ES index.  So the original assertion is confirming that by scanning across the search results returned.

From the diff, it looks like you've modified the test slightly (for readability?) but as a side effect it is now checking the wrong key. Seems like should be checking that the search result was returned from the index matching`index_alias_language`?  Since your change, the value it is checking `index_alias` just has the name of the default ES index.     If we only found the key in that index, it would indicate that expected behavior failed (we did not successfully store into the appropriate language specific index).  Or, the behavior of the underlying code has changed and we are no longer surfacing results from multiple indexes?

DGaffney:
> > this looks to be failing just on the language test, but at first blush, I don't even know if this is really a failure mode? It would seem to me that this could be an entirely intended end-state of this test, no?
> 
> @DGaffney The intention of the test is to validate the behavior that text of one of the recognized languages is provided, it is added to the appropriate language specific ES index. So the original assertion is confirming that by scanning across the search results returned.
> 
> From the diff, it looks like you've modified the test slightly (for readability?) but as a side effect it is now checking the wrong key. Seems like should be checking that the search result was returned from the index matching`index_alias_language`? Since your change, the value it is checking `index_alias` just has the name of the default ES index. If we only found the key in that index, it would indicate that expected behavior failed (we did not successfully store into the appropriate language specific index). Or, the behavior of the underlying code has changed and we are no longer surfacing results from multiple indexes?

The error message here indicates that we expected `alegre_similarity_test` to be present in the list `['alegre_similarity_test_hi']` which, if that's the result list of indices, wouldn't that mean that it _did_ send the document to the correct language specific index? The change here is that it is not present in `alegre_similarity_test` but is present in `alegre_similarity_test_hi` which seems...... fine?

skyemeedan:
> The error message here indicates that we expected `alegre_similarity_test` to be present in the list `['alegre_similarity_test_hi']` which, if that's the result list of indices, wouldn't that mean that it _did_ send the document to the correct language specific index? The change here is that it is not present in `alegre_similarity_test` but is present in `alegre_similarity_test_hi` which seems...... fine?

So it needs to check with the key `index_alias_language` which should have the value `'alegre_similarity_test_hi'` and the test will pass?.   Currently it is checking with `index_alias`   which has the value `'alegre_similarity_test'`

DGaffney:
> > The error message here indicates that we expected `alegre_similarity_test` to be present in the list `['alegre_similarity_test_hi']` which, if that's the result list of indices, wouldn't that mean that it _did_ send the document to the correct language specific index? The change here is that it is not present in `alegre_similarity_test` but is present in `alegre_similarity_test_hi` which seems...... fine?
> 
> So it needs to check with the key `index_alias_language` which should have the value `'alegre_similarity_test_hi'` and the test will pass?. Currently it is checking with `index_alias` which has the value `'alegre_similarity_test'`

Yes what I'm implying here is that `alegre_similarity_test` is an optional index to return while `alegre_similarity_test_{lang}` would be required. @computermacgyver please weigh in when you have a moment!

computermacgyver (approved):
Updated test on language-specific indices looks good to me. Thanks for doing this!

skyemeedan (approved):

PR #490, opened 2025-02-13
[CV2-6066] load tests for text similarity endpoint

skyemeedan:
## Description

* adding some comments and questions in the code for later refactoring
* adding a benchmark script (adapted from timpani benchmarks) that can validate API contract and generate large volumes of text similarity queries for testing.  C

Validates some contracts
runs 100 item test in a basic loop using the old, async, and sync endpoints.  FROM A SINGLE PROCESS
runs 100 item test for old and new blocking endpoints 10 ITEMS EACH FROM 10 PROCESS

reports some timings (I'm not sure I trust these at all for the parallelization case, as I see them printed while queries still showing up in logs an process has not finished).  

TODO: maybe unitest framework messing with parallelization, call from CLI 

Reference: https://meedan.atlassian.net/browse/CV2-6066

## How has this been tested?
I have run this code locally on the alegre container.  Need to ensure that the old alegre paraphrase multilingual and new presto paraphrase multilingual models are running    

`python app/test/benchmark/text_similarity_benchmark.py --verbose -f`

I've confirmed that the tests do not get run by the CI tests (which is why they are in benchmark dir, and file does not have 'test' in name) 

Would be good to make it so that we can't run this script in live unless we *really* mean to.

### DIFF
--- a/app/test/benchmark/text_similarity_benchmark.py
+++ b/app/test/benchmark/text_similarity_benchmark.py
@@ -8,17 +8,15 @@ from concurrent.futures import ThreadPoolExecutor
 import faker
 
 
-class AlegreTextAPILoadTest(unittest.TestCase):
+class AlegreTextSimilarityAPILoadTest(unittest.TestCase):
     """
-    Not sure where this test belongs. Should not be part of
-    normal CI and unit tests. Goal is to test API limits
-    of functionality in Alegre that Timpani will need.
+    Load test Alegre text similarity endpoints for benchmark tests
+    and debugging load failure states
     """
 
     LOCALES = ["es", "en_US", "zh_CN", "ja_JP", "pt", "or_IN", "hi_IN"]
     fake_content_source = faker.Faker(LOCALES)
 
-    # TODO: this is for local, need to change for QA
     # http://alegre.qa but probably can only run when deployed
     # TODO: don't run this in live
     # ALEGRE_BASE_URL = "http://alegre:3100"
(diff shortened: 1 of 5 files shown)


app/test/benchmark/text_similarity_benchmark.py
```diff
@@ -0,0 +1,446 @@
...
+    """
+    Load test Alegre text similarity endpoints for benchmark tests
+    and debugging load failure states
+    """
+
+    LOCALES = ["es", "en_US", "zh_CN", "ja_JP", "pt", "or_IN", "hi_IN"]
+    fake_content_source = faker.Faker(LOCALES)
+
+    # http://alegre.qa but probably can only run when deployed
+    # TODO: don't run this in live
```
caiosba, line 21: Is there a way to guard against that?

caiosba (commented):
There is a lot of clean up that can go into `app/test/benchmark/text_similarity_benchmark.py` (e.g., do we still need those comments from August 2023? is the code the right place to document those results?), but that can happen later, since this is not production code.

app/test/benchmark/text_similarity_benchmark.py
computermacgyver: Of course:
```python
import os

if os.environ.get("DEPLOY_ENV")!= "qa":
   exit("This script can only be run in QA")
```

computermacgyver (approved):

PR #494, opened 2025-02-26
[CV2-5885] similarity overview docs

skyemeedan:
## Description
Adds some additional documentation files and diagrams describing the relationships between observable parts of Alegre infrastructure.  The diagrams reproduce the 'service overview' created by Devin, but include redis and 3rd party API information

Also renames and updates some previous documentation, moves documentation images into doc folder, and links the new documentation from README

NOTE: in github, choose "View File" to view the rendered markdown with diagrams

Reference: https://meedan.atlassian.net/browse/CV2-5885

## How has this been tested?
The diagrams render correctly as markdown


caiosba (approved):
Looks great, thanks Skye! My suggestion is to make clear that the Alegre Bot is nothing more than just a module inside Check API. The way it is right now feels like it's independent from Check API, which is not the case. Let's talk more about it in our 1:1.

skyemeedan:
> Alegre Bot is nothing more than just a module inside Check API.

If that is the case, I'll update it to just say CheckAPI.   Is Timpani the only thing that calls Alegre without going through CheckAPI?  What about the indexing rake tasks?

caiosba:
> > Alegre Bot is nothing more than just a module inside Check API.
> 
> If that is the case, I'll update it to just say CheckAPI. Is Timpani the only thing that calls Alegre without going through CheckAPI? What about the indexing rake tasks?

Yes, AFAIK, Timpani is the only other Alegre client besides Check API. The indexing rake tasks are also inside Check API, so I think it's OK to keep only Timpani and Check API.

PR #499, opened 2025-03-04
[CV2-6075] text similarity sync & async call diagrams

skyemeedan:
## Description
Adds markdown documentation files for tracing through the text similarity calls.  Although I believe these are accurate (they were written by tracing through the code) it would be great to take another pass through at confirming details

Reference: https://meedan.atlassian.net/browse/CV2-6075

### DIFF
--- a/doc/text-similarity-call-sequence.md
+++ b/doc/text-similarity-call-sequence.md
@@ -1,4 +1,4 @@
-# Sequence diagram for text similarity calls
+# Sequence diagram for text similarity calls [DRAFT]
 
 There are multiple paths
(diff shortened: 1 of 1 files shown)


caiosba (commented):
There is a typo in `recieved`, and the questions in the documentation are pertinent. The Check API part is accurate AFAICT. I suggest the title of the diagram states `[DRAFT]` explicitly, but looks like it's almost there. I appreciate if @computermacgyver can confirm from the Alegre/Presto side.

caiosba (approved):

PR #502, opened 2025-04-14
Don't start model servers on live.

caiosba:
## Description

On the Check API side, all requests are routed to Presto, so we don't need to have the Alegre model servers running, as they are already managed by Presto.

Reference: CV2-6259.

## How has this been tested?

After deployment, we need to make sure that the Alegre model services don't start.

N/A.


caiosba:
@sonoransun can you please confirm this is all we need to NOT start the model services on deployment, for both QA and live?

sonoransun (approved):
This should do it!

sonoransun:
> @sonoransun can you please confirm this is all we need to NOT start the model services on deployment, for both QA and live?

We'll also need to turn down the instance counts, but that can be done when ready.

PR #505, opened 2025-08-04
Trying to fix build error on check repo

caiosba:
I noticed that the `check` build was installing a different version of `tokenizer`, despite the pinned version in `requirements.txt` (maybe it was coming from another dependency). That different version was triggering an error on `check`'s CI. This PR ensures that the version we need is installed.

I also took the opportunity to fix some failing tests.

Reference: CV2-6381.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN echo "set enable-bracketed-paste off" >> ~/.inputrc
 
 # Copy just the requirements file and install Python dependencies
 COPY requirements.txt ./
+COPY constraints.txt ./
 RUN pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


danielevalverde (approved):
Thanks, @caiosba. This change appears to have resolved the Alegre build issue on the Check repository.
Before your changes, the build was consistently hanging during the pip install step for  Alegre and queue_worker and  now this step was completed without interruption

`#88 [queue_worker  9/11] RUN pip install --no-cache-dir -r requirements.txt
#88 DONE 290.8s
#136 [alegre  9/11] RUN pip install --no-cache-dir -r requirements.txt
 `
 Log here: https://github.com/meedan/check/actions/runs/16757125892/job/47442589617

